PR #7, opened 2020-08-03
Google sheets improvement

JasonLin43212:
- Dynamically get the values from the Google sheet using a range value of `Sheet1!A:D`. However, this breaks if a row is missing some values.
- Converted each row from a list into a dictionary where the key is the header value (title, front_src, ...) and the value is the value associated with that header. (There doesn't seem to be a way for the Google Sheets API to return a dictionary)

- For cleaning up the authentication pickling routes, we can just have:
```python
flow = InstalledAppFlow.from_client_secrets_file(
        settings.GOOGLE_API_CREDENTIALS_FILE,
        SCOPES
)
creds = flow.run_local_server(port=8080)
```
and everything works fine, but a new window pops up asking us to choose a Google account every time we run `python manage.py syncdb`

- I'm not sure if it is just me, but I am able to sync the db even when my back end is running.

### DIFF
--- a/backend/app/management/commands/syncdb.py
+++ b/backend/app/management/commands/syncdb.py
@@ -48,9 +48,7 @@ class Command(BaseCommand):
 
         print_header(f'Will import range {spreadsheet_range}')
 
-        # TODO(ra): clean up authentication pickling routines --
         # Settings for pickle file
-        # do we even want to cache auth to disk? probably not...
 
         creds = None
         # The file token.pickle stores the user's access and refresh tokens, and is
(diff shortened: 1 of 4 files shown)


backend/app/management/commands/syncdb.py
```diff
@@ -2,9 +2,7 @@
 Django management command syncdb
 
 Syncs local db with data from project Google Sheet
-
-TODO(ra): link Google Sheet here
-"""
+s"""
```
ryaanahmed, line 5: Not blocking a merge, but just noting this errant `s` -- I know your next work is going to involve this file, so cleanup then.

ryaanahmed:
> Dynamically get the values from the Google sheet using a range value of Sheet1!A:D. However, this breaks if a row is missing some values.

This is okay for now, but let's fix quickly bc I suspect this is going to bite us early in prototyping. Probably just need to replace this kind of thing `row["title"]` with `row.get('title', '')` so that we can provide a default value.

ryaanahmed:
Also we'd like to get to the point where we don't have to explicitly set the row range A:D

PR #8, opened 2020-08-04
Google Sheets Foreign Keys

JasonLin43212:
- Added MapSquare and Photographer to the Google Sheets and created a way to add the information into the database with syncdb
- Syncdb now automatically gets all rows and columns without having to specify a range.
- Syncdb will not break when a row contains missing values. Those values will just be left blank.
- Removed ManyToManyField Photo field from MapSquare and Photographer models
--Photos in both serializers now retrieved by filtering photo objects with matching instance id


backend/app/management/commands/syncdb.py
```diff
@@ -92,21 +95,27 @@ def handle(self, *args, **options):
         # THIS IS JUST FOR PROTOTYPING NEVER EVER EVER EVER IN PRODUCTION do this
         superuser = User.objects.create_superuser('admin', password='adminadmin')
 
-        if not values:
+        if not databases:
```
ryaanahmed, line 98: In cases like this where there's an exit condition and then you run all of the business logic otherwise, my preference is this style (once it gets long enough... which it has at this point!):

```
if not databases:
    print_header('No data found')
    return

# the rest of your code goes here...
```

backend/app/management/commands/syncdb.py
```diff
@@ -92,21 +95,27 @@ def handle(self, *args, **options):
...
+                    if model_name == 'Photo' or model_name == 'Photographer':
+                        map_square_name = row.get('map_square_obj', '')
+                        # Returns the object that matches or None if there is no match
+                        row['map_square_obj'] = \
+                            MapSquare.objects.filter(name=map_square_name).first()
+                    if model_name == 'Photo':
+                        photographer_name = row.get('photographer_obj', '')
+                        row['photographer_obj'] = \
+                            Photographer.objects.filter(name=photographer_name).first()
+                    model_instance = eval(f'{model_name}(**row)')
```
ryaanahmed, line 120: No `eval`! I'm surprised that's not in our pylint settings; I'll double check. (The basic problem with `eval` is that because you're passing code as a string, it doesn't get linted, typechecked, or checked for syntax before running.)

 You can do this in code by saving the correct model class to a variable instead:
```
if model_name == 'Photo':
    ModelClass = Photo
# ... do other stuff ...
model_instance = ModelClass(**row)
model_instance.save()
```

backend/app/management/commands/syncdb.py
```diff
@@ -92,21 +95,27 @@ def handle(self, *args, **options):
...
-            for row in values_as_a_dict:
-                print(row)
-                photo = Photo(
-                    title=row["title"],
-                    front_src=row["front_src"],
-                    back_src=row["back_src"],
-                    alt=row["alt"],
-                )
-                photo.save()
+            for m, values in enumerate(databases):
```
ryaanahmed, line 101: you can use `zip` here:

`for model_name, values in zip(spreadsheet_ranges, databases):`

and then you don't have to pass the index to get the model_name

backend/app/management/commands/syncdb.py
```diff
@@ -44,9 +44,9 @@ def add_arguments(self, parser):
         parser.add_argument('--range', action='store', type=str)
 
     def handle(self, *args, **options):
-        spreadsheet_range = 'Photo!A:D'
+        spreadsheet_ranges = ['MapSquare', 'Photographer', 'Photo']
```
ryaanahmed, line 47: It looks to me like in order for this to run correctly, this order is important. Please add a comment here specifying this, so later readers don't accidentally change the order values.

backend/app/models.py
```diff
@@ -27,10 +27,6 @@ class MapSquare(models.Model):
     the boundary of this Map Square.
     """
     name = models.CharField(max_length=252)
-    photo_objects = models.ManyToManyField(
```
ryaanahmed, line 30: Thanks!

backend/app/models.py
```diff
@@ -43,8 +39,4 @@ class Photographer(models.Model):
     name = models.CharField(max_length=252)
     type = models.CharField(max_length=252)
     sentiment = models.CharField(max_length=252)
-    photo_objects = models.ManyToManyField(
-        'Photo',
-        blank=True,
-    )
     map_square_obj = models.ForeignKey('MapSquare', on_delete=models.SET_NULL, null=True)
```
ryaanahmed, line 42: Before we get too deep into this project, let's rename these just `map_square` and wherever `photographer_obj` is just call it `photographer`. Typing `_obj` all the time for these is going to get tedious.

PR #12, opened 2020-08-06
updated photoview

michellejhe:
created a basic photo view that matches Jason's wireframe

### DIFF
--- a/frontend/src/photoView/photoView.js
+++ b/frontend/src/photoView/photoView.js
@@ -77,9 +77,9 @@ export class PhotoView extends React.Component {
                      style={{ padding: '100px 50px', float: 'left', marginTop: 50 }}>
                      <div className='attributeTitle' style={{ fontWeight: 'bold' }}>
                         Photographer:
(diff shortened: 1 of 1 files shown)


frontend/src/photoView/photoView.js
```diff
@@ -28,35 +29,105 @@ export class PhotoView extends React.Component {
...
+                 </div>
+                 <div className='imageInfo'
+                     style={{ padding: '100px 50px', float: 'left', marginTop: 50 }}>
+                     <div className='attributeTitle' style={{ fontWeight: 'bold' }}>
+                        Photographer:
+                         <div className='attribute'style={{ fontWeight: 'normal' }}>
+                            None
+                         </div>
+                     </div>
+                     <div className='attributeTitle' style={{ fontWeight: 'bold' }}>
```
ryaanahmed, line 84: I think _probably_ this repeated code is a sign that we need to refactor a bit. But let's take a look together in our call -- this is already good enough to get us to where we want to get with the prototype today!

PR #13, opened 2020-08-07
Integrate Google Drive API to show images

JasonLin43212:
- When running syncdb, it will use the map square number and the photo source to find the url from Google Drive and put that link into the database.

NOTE: Merge `improving_syncdb` first since this branch builds on top of what Justice did.


ryaanahmed:
Also worth noting here: first time I ran `syncdb` on this branch, I got the following error:

`googleapiclient.errors.HttpError: <HttpError 403 when requesting https://www.googleapis.com/drive/v3/files?q=%271aiY1nFJn6T7khu5dhIs3U2o8RdHBu6V7%27+in+parents&fields=nextPageToken%2C+files%28id%2C+name%29&alt=json returned "Insufficient Permission: Request had insufficient authentication scopes.">`

This is because I had the pickled token file that we generate from when we only had _one_ scope. Deleting that and rerunning resolves the issue, so just FYI @dhmit/urop-2020-summer, make sure to delete your `backend/token.pickle` after this goes in.

I don't think we need to do anything fancy like check to see whether the token has sufficient creds and request another if not bc I don't think we're going to be updating the scopes very frequently (if at all after this). But if that changes, let's keep this in mind so that we don't crash the team whenever we roll out something that calls a new part of the Google API.

PR #14, opened 2020-08-07
Improving syncdb

JusticeV452:
- Added shelfmark, contains_sticker, librarian_caption, photographer_caption to Photo
- Removed title field from Photo
- Added number field to all models
- References to other models changed from id to "number"
- syncdb now resets migrations
- Changing a model's fields does not break syncdb


backend/app/management/commands/syncdb.py
```diff
@@ -113,16 +120,49 @@ def handle(self, *args, **options):
...
+                    if header in model_field_names or header == 'map_square_number':
+                        # Check if value in column is a number
+                        value = row[header]
+                        if header in ['number', 'map_square_number', 'photographer']:
+                            try:
+                                value = int(value)
+                                if header == 'map_square_number':
+                                    header = 'map_square'
+                            except ValueError:
+                                continue
```
ryaanahmed, line 137: In cases like this, let's print to the console a note that the user should go check the value in the spreadsheet, since there's a mistake in the metadata.

backend/app/management/commands/syncdb.py
```diff
@@ -88,12 +88,19 @@ def handle(self, *args, **options):
             values = result.get('values', [])
             databases.append(values)
 
+        # Delete database
         if os.path.exists(settings.DB_PATH):
             print_header('Deleting existing db...')
+            for file in os.listdir(settings.MIGRATIONS_DIR):
```
ryaanahmed, line 94: Part of the idea behind having the migration history is that if you want to rollback a change to your database models, the history of files in the migration directory gives you a way to do that. Clobbering the whole directory, as we do here, removes that feature.

Buuuuut... I'm okay with this for now. In early prototyping, we're often making a lot of quick changes to the database schema, and no one's reliant on this being stable. Let's open an issue to _remove_ this feature at the end of the summer, once the models are in a more stable state.

backend/app/management/commands/syncdb.py
```diff
@@ -113,16 +120,49 @@ def handle(self, *args, **options):
 
             for row in values_as_a_dict:
                 print(row)
+                # Filter column headers for model fields
+                model_fields = MODEL_NAME_TO_MODEL[model_name]._meta.get_fields()
```
ryaanahmed, line 124: nice work here!

backend/app/management/commands/syncdb.py
```diff
@@ -113,16 +120,49 @@ def handle(self, *args, **options):
...
+                        if header in ['number', 'map_square_number', 'photographer']:
+                            try:
+                                value = int(value)
+                                if header == 'map_square_number':
+                                    header = 'map_square'
+                            except ValueError:
+                                continue
+                        # Evaluate value as a boolean
+                        elif header == 'contains_sticker':
+                            if value.lower() == 'yes':
```
ryaanahmed, line 140: This is cool, but I think we should just agree on a way of coding this data in the sheet, and handle similarly to the above ValueError case in case of incorrect coding.

PR #24, opened 2020-08-10
Fixed Navbar Overlapping Content?

### LINKED ISSUE #19: Handle various photo aspect ratios
Right now, we force everything into a fixed width/height.

JasonLin43212:
- Added a className `page` that is used by all top level containers to add enough margins to make the navbar not overlap with the content.
- Tidy up photoView CSS so that it is a bit responsive and the image's aspect ratio is preserved. (I just realized that this also fixes #19 )


frontend/src/index/index.scss
```diff
@@ -18,12 +18,12 @@ code {
   margin-bottom: 1px;
 }
 
-.landing-page {
+.page {
     padding: 55px 55px 120px;
```
ryaanahmed, line 22: SASS allows us to define variables, and when we're doing things like clearing a header or footer it's very useful to, e.g., set a `$header-height` var instead of using magic numbers. That'll make this stylesheet more readable/usable for future maintainers.

frontend/src/photoView/photoView.scss
```diff
@@ -0,0 +1,11 @@
+.image-photo {
```
ryaanahmed, line 1: Thanks for factoring this out.

ryaanahmed:
Thanks! Merging.

PR #25, opened 2020-08-10
Photographer Page

JasonLin43212:
- Created a photographerView to display information about a photographer
- Made the path `/photographer/<int>/` link to the photographer view page.
- Fixed serializer for photographers


frontend/src/photographerView/photographerView.scss
```diff
@@ -0,0 +1,27 @@
+.photographer-name {
```
ryaanahmed, line 1: Rather than manually set each of these as a different font-size in the CSS, we should use semantic HTML. e.g., `<div className='photographer-name'>` above should probably be an `<h1>`, the individual things like attributes could be `<h2>` etc.

See: https://webaim.org/techniques/semanticstructure/ for why -- it's both good for accessibility (lets devices like screen readers understand the structure of the page) and means we can setup universal heading styles across the project if we don't like the built-in bootstrap styles.

ryaanahmed: Same deal applies to the attributes in the PhotoView, btw.

ryaanahmed:
thanks! merging.

PR #26, opened 2020-08-10
Catching syncdb permission error

JusticeV452:
- syncdb catches permission error and returns a helpful error message
- syncdb now attempts to the delete database first to avoid unnecessary waiting


ryaanahmed:
Thanks -- works great.

PR #27, opened 2020-08-12
Variable images

JasonLin43212:
- Updated Google Sheets, models, serializers, and syncdb to handle front, back and binder src dynamically.
- Display the different sides of the photo dynamically with buttons that only let you flip to a side if it is available.
- Fixed some key errors and header tag errors


frontend/src/photoView/photoView.js
```diff
@@ -2,14 +2,16 @@ import React from 'react';
 import Navbar from '../about/Navbar';
 import { Footer } from '../UILibrary/components';
 
+const SIDES = ['front', 'back', 'binder'];
```
ryaanahmed, line 5: Ideally what we'd like to do for a situation like this is use an enum, which JavaScript doesn't really natively support (one of the many reasons to love [TypeScript](https://www.typescriptlang.org/docs/handbook/enums.html)) .

Here's the workaround -- use an object. So:
```
const SIDES = {
    FRONT: 'front',
    BACK: 'back',
    BINDER: 'binder',
}
```

Then, when later you're switching on which side, instead of switching on a string like this: `if (this.state.display_side === 'front')`

you can do `if (this.state.display_side === SIDES.FRONT)`

Since you're no longer using a string literal but instead a variable, this is now much more robust to things like typos.

ryaanahmed:
thanks for the changes -- merging.

PR #28, opened 2020-08-12
Added Map Square View

JasonLin43212:
- Added Map Square View
  - Displays the map square name and a list of the photos that belong to that Map Square
  - Updated the homepage to link the map square to its respective page


frontend/src/mapSquareView/mapSquareView.js
```diff
@@ -0,0 +1,69 @@
...
+                });
+            }
+        } catch (e) {
+            console.log(e);
+        }
+    }
+
+
+    render() {
+        if (this.state.loading) {
```
ryaanahmed, line 34: This "loading" infrastructure is starting to look like something we can factor out into its own component.

ryaanahmed:
looks good to me!

PR #31, opened 2020-08-14
Use the Photo number to find the photo URL

JasonLin43212:
- Instead of having a column with all the sources, we now use the number column to dynamically find the Google Drive url of a photo


backend/app/management/commands/syncdb.py
```diff
@@ -209,21 +220,13 @@ def handle(self, *args, **options):
...
-                            # the person who entered the data added extra spaces
-                            photo_srcs.append(src.strip())
-                    else:
-                        photo_srcs.append(srcs)
                     map_square_folder = photo_url_lookup.get(map_square_number, '')
+                    photo_number = row.get('number', '')
 
                     if map_square_folder:
-                        add_photo_srcs(model_kwargs, map_square_folder, photo_srcs)
+                        add_photo_srcs(model_kwargs, map_square_folder, str(photo_number))
```
ryaanahmed, line 229: This is so absurdly nitpicky that I'm certainly not going to request changes for it -- but probably the better thing here is for the function to take `photo_number` as an int (since that's how it mostly will exist in the calling environment) and to do the cast in the function itself. NOT A BIG DEAL and purely a judgement call... but if this were in a situation where you'd be reusing `add_photo_srcs` a lot, it would be come inconvenient and error prone for the caller to always be responsible for that cast.

ryaanahmed:
Great -- this substantially simplifies and reduces the possible sources of error as we transcribe metadata.

PR #32, opened 2020-08-14
Pickling Analysis

JasonLin43212:
- Added an optional argument called `rerun` which you input after the `analysis_name`. You must use "rerun" for the second argument in order for the analysis to rerun on all instances in the Model
- Added a field to each analysis file that specifies the Model name (we might want to have a new model for more general analysis?)
- Stores the analysis results in a pickle file that is located in `/analysis/analysis_results` and uses past results when you are not rerunning

NOTE: Contains Justice's white space analysis


JasonLin43212:
I'm not sure what is going on with the python lint error.

ryaanahmed:
Ach, sorry -- see my comment over in #33. It's because we removed the whole directory that the linter is looking for (the old top-level analysis dir). I'll fix, don't worry about it. Just need to change the GitHub Actions config file.

ryaanahmed (approved):
Looks good. I'm not sure about one design decision here: this PR lifts the `Model.objects.all()` out of the individual analysis functions into the calling code in `runanalysis`.

We may find down the road that we'll want to do analyses not on the entire set of data for a given model, but on a filtered queryset. (e.g., an analysis that only runs on Photo objects filtered by a certain other parameter), so my instinct is to push the query back into the individual analyses.

I'll merge as is now and have a proper think about this before making any additional changes.

PR #33, opened 2020-08-14
Calculate whitespace ratio

JusticeV452:
- Calculates the ratio of pixels with a value >= 60% of the brightest pixel in the image, to the image size
- Retrieves photos using the public google drive download
- Currently breaks after a retrieving too many photos due to rate limit (HTTP 403 error)
  - Ryaan will look into a way of configuring the developer console to allow authorized file requests that can be used instead of retrieving the photos via the public link.


ryaanahmed:
Whoops -- the linter error here is my bad: we removed that analysis folder at the project root, but there's still a GitHub CI job that's looking for it! I'll fix.

ryaanahmed:
Merging now despite linter failure -- I've got to do some settings work on that, so will fix up and make a separate PR with lints.

PR #37, opened 2020-08-18
Added a Corpus Analysis model

JasonLin43212:
- Added the `CorpusAnalysisResult` model along with a serializer, a view, and an API endpoint to get all corpus analysis results.


ryaanahmed (approved):

ryaanahmed:
Looks good. I think inspired by this, I'm actually going to make a more general `AnalysisResult` model. I'm realizing that as folks make more and more analyses, it's going to become onerous to keep adding fields to the database models. Instead, we'll just have a generic `AnalysisResult` that can take arbitrary data serialized as JSON, just like this one does, and point them via foreign key to the model instances they belong to.

PR #61, opened 2020-09-25
Fixed bug where get_image_data wouldn't use local cleaned path

L-Jewel:
This PR:
- Fixes an if-statement within the get_image_data() function


ryaanahmed (approved):
Thanks for the quick fix!

PR looks great, but notice I changed the title: you want the PR title to be descriptive of the general scope of the change. Saying you fixed an if statement somewhere doesn't help the reviewer quickly see the context/purpose of the PR... and that'll be evident when the reviewer looks at the `Files changed`.

PR #64, opened 2020-10-05
Created AnalysisView

JusticeV452:
- Shows small photos in order of analysis result value at url analysis/<analysis_name>.
- Photos link to photo page and display analysis result value on hover.


ryaanahmed:
@JusticeV452 -- ach, somehow accidentally merged this instead of requesting changes! The intention behind the TextField on the AnalysisResults class was to have everything JSON serialized, and this PR removes that. The rest of the infrastructure is great, but I'm going to hotfix master to remove that change. We'll have to implement deserializing and sorting in the backend to make this work properly.

ryaanahmed:
@JusticeV452 -- No, scratch that. It looks like GitHub somehow got confused -- I merged a completely separate PR (maybe with the same branch name? did you share between two projects?) and for some reason that triggered it identifying this PR as having been merged, even though the commit https://github.com/dhmit/paris_1970/pull/64/commits/27472d5c6f7bc15aeaad883ee10e76b49813eb39 has not been merged into the master branch. Super weird; I've never seen this before.

Anyway: above comment still stands. The frontend work here is great, but the backend changes break the intention to JSON serialize more complex analysis results. E.g., groups that have analyses that produce bounding boxes can currently serialize lists or dicts as JSON, which we might want to pass to the frontend to display.

Let's do the following:
- [ ] Restore `AnalysisResult.result` to a `TextField`
- [ ] In the serializer, rather than doing a `order_by`, deserialize the JSON results field and sort by that

I can't seem to reopen this PR, let's continue work on this branch and open a new PR with these changes.

PR #66, opened 2020-10-09
Detail Analysis Functions

jlin00:
PR:
- detail.py: uses a fast Fourier transform to analyze the amount of sharp changes in a photo
- local_variance.py: uses the Laplacian operator to analyze how sharp the edges are in a photo
- stdev.py: isolates local regions of an image and finds the average of all standard deviations to determine
level of detail in a photo
- average_detail.py: takes the average of the numerical values returned by the previous three functions to compute the average detail in a photo


backend/app/analysis/tests.py
```diff
@@ -32,19 +36,117 @@ def setUp(self):
...
+                self.photo_dict[photo_name] = photo
 
     def test_photographer_caption_length(self):
-        self.photo_0.photographer_caption = '123456'
-        result = photographer_caption_length.analyze(self.photo_0)
-        self.assertEqual(6, result)
+        """
+        Test Photographer Caption Length analysis (photographer_caption_length.py)
+        """
+        expected_values = {'100x100_500px-white_500px-black': 6}
```
ryaanahmed, line 57: The use of the `expected_values` dict, the for loop, and the `if image in self.photo_dict.keys()` are all unnecessary here: we know exactly which image we're looking at, so there's no need to obscure that with a for loop that iterates over a collection of only one object, etc. Just define the key you're calling into `photo_dict` and the result you expect explicitly, and remove those extra bits of flow control.

`if image in self.photo_dict.keys():` is particularly problematic here -- if for some reason our code up in `setUp` has failed and this key _isn't_ in the dict as expected, we want this test to crash. Presently, it will silently pass without doing anything because the block inside the `if` won't execute. These kinds of guardrails are sometimes useful in production code, but in testing code, you want to avoid situations where you write around crashes.

backend/app/analysis/tests.py
```diff
@@ -32,19 +36,117 @@ def setUp(self):
...
+                print(f'Caption Length performed on {image}. Result: {result}')
+                self.assertEqual(expected_values[image], result)
 
     def test_whitespace_percentage(self):
-        result = whitespace_percentage.analyze(self.photo_0)
-        self.assertEqual(50, result)
+        """
+        Test Whitespace Percentage analysis (whitespace_percentage.py)
+        """
+        expected_values = {'100x100_500px-white_500px-black': 50}
```
ryaanahmed, line 69: same as `test_photographer_caption_length` -- no need for the for loop or the if.

backend/app/analysis/tests.py
```diff
@@ -32,19 +36,117 @@ def setUp(self):
...
+
+    def test_stdev(self):
+        """
+        Test Standard Deviation analysis (stdev.py)
+        """
+        expected_values = {'100x76-BlackRectangle': 0, '100x76-GreyRectangle': 0,
+                           '100x76-WhiteRectangle': 0, '100x100-BlackSquare': 0,
+                           '100x100-GreySquare': 0, '100x100-WhiteSquare': 0,
+                           '100x76-CheckeredRectangle_1': 126, '100x76-CheckeredRectangle_2':
+                               126, '100x76-HalfRectangle_1': 1, '100x76-HalfRectangle_2': 126,
```
ryaanahmed, line 84: nitpick -- to make this more readable, don't put line breaks after colons.

not...
```
'100x76-CheckeredRectangle_1': 126, '100x76-CheckeredRectangle_2':
    126, '100x76-HalfRectangle_1': 1, '100x76-HalfRectangle_2': 126,
```

but...
```
'100x76-CheckeredRectangle_1': 126, '100x76-CheckeredRectangle_2': 126,
'100x76-HalfRectangle_1': 1, '100x76-HalfRectangle_2': 126,
```

backend/app/analysis/tests.py
```diff
@@ -32,19 +36,117 @@ def setUp(self):
...
+                           '100x76-HalfRectangle_3': 1, '100x76-HalfRectangle_4': 126,
+                           '100x100-CheckeredSquare_1': 125, '100x100-CheckeredSquare_2': 125,
+                           '100x100-HalfSquare_1': 0, '100x100-HalfSquare_2': 125,
+                           '100x100_500px-white_500px-black': 0,
+                           '100x100-HalfSquare_4': 125}
+
+        for image in expected_values:
```
ryaanahmed, line 91: here the for loop makes perfect sense, but the `if` has the same problem: if the image isn't there, we want this test to crash because something has gone wrong.

backend/app/analysis/tests.py
```diff
@@ -32,19 +36,117 @@ def setUp(self):
...
+                           '100x76-CheckeredRectangle_1': 3087, '100x76-CheckeredRectangle_2':
+                               3087, '100x76-HalfRectangle_1': 1717, '100x76-HalfRectangle_2': 1305,
+                           '100x76-HalfRectangle_3': 1717, '100x76-HalfRectangle_4': 1305,
+                           '100x100-CheckeredSquare_1': 2660, '100x100-CheckeredSquare_2': 2660,
+                           '100x100-HalfSquare_1': 1305, '100x100-HalfSquare_2': 1305,
+                           '100x100_500px-white_500px-black': 1300,
+                           '100x100-HalfSquare_4': 1305}
+
+        for image in expected_values:
+            if image in self.photo_dict.keys():
```
ryaanahmed, line 113: same deal with the `if`

ryaanahmed: and another couple below -- you'll see them

backend/app/analysis/stdev.py
```diff
@@ -0,0 +1,59 @@
...
+
+    # Declare and initialize boundary markers to split flatList into segments
+    i = 0
+    j = flat_list.size // 64
+    interval = flat_list.size // 64
+
+    while j <= flat_list.size:
+        # Append segments of the flattened grayscale image array to localRegions
+        local_regions.append(flat_list[i: j])
+        i, j = j, j + interval
```
ryaanahmed, line 46: (nitpick) I avoid multiple assignment because it tends to make code less legible.

This...
```
i = j
j = j + interval
```
... is clearer to read. Here, not such a big deal, but in more complex situations it can become a problem for collaborators.

ryaanahmed (changes requested):
Really good work! I have some comments about changes to the testing system that are the biggest things that need doing here before merge.

The clarity and comments in the actual analysis code are excellent! Very easy to read and straightforward to see what's going on in each function.

Please add comments in your individual analysis functions and also your aggregate `average_detail` function giving the expected range of outputs.

Finally: throughout this PR we use the word `average` loosely. Good to specify that we mean the `mean` throughout, since this kind of analysis code is for sure a place where we might end up using a different kind of average elsewhere.

ryaanahmed (approved):

ryaanahmed:
Thanks for making all the requested changes. Looks good! Merging.

PR #67, opened 2020-10-16
Indoor analysis using three distinct methods

montsegarza:
The indoor analysis determines if an image was likely taken indoors or outdoors. It has four new analysis functions inside an indoor_analysis folder for better organization. All three functions return booleans. They are:

1. Courtyard Frame Analysis - looks at the border of an image to determine whether it was framed within an alleyway like most courtyard images are. Returns true if frame is present, which indicates it is likely an indoor image
2. Gradient Analysis - looks at the brightness gradient on an image top-down to determine if a sky is present, which would indicate the image is outdoors. Returns true if a sky is present, indicating image is likely outdoors.
3. Find Window Analysis - using the good features to track function in the OpenCV library, this analysis finds perfect rectangles in an image indicating the possibility of a window being present and thus, the image having been taken outside. Returns true if possible window found.
4. Combined Analysis - Using the three methods above, returns true if two out of the three methods deem the image to have been taken indoors. Returns false otherwise

Many tests have also been added in tests.py. Images for our tests are organized in folders within the test_images directory.

TODO:
- [ ]  Further test the functions to see where they fail and how they can be improved.
- [ ]  Consider more methods to identify if an image is indoors or outdoors.


ryaanahmed (changes requested):
Reviewed over call -- requesting some small changes for readability.

montsegarza:
@ryaanahmed I think we have fixed everything you mentioned on the call

ryaanahmed:
Thanks for the changes! Looks good -- merging.

Please note that #66 introduced a more concise method for adding test Photo instances. @montsegarza @YifanWang0 @radha-patel -- when you have a chance, please take a moment to clean up your tests to use the new system.

ryaanahmed:
Actually -- turns out the new method for adding Photos needs a little extension to handle photos in subdirectories, as this PR adds. I'll go ahead and do the cleanup myself.

ryaanahmed (approved):

PR #69, opened 2020-10-16
Photo OCR

JasonLin43212:
The `text_ocr` function takes in an image and outputs a string of the text that appears in the image. This uses the OpenCV's EAST text detector and Tesseract to detect and then recognize the text in the image. We included tests for the function but it seems to only work well with user created tests (i.e. an image of a white background with black text in the middle).

For the photos from the database, we encountered many obstacles with the text recognition. For example, photos in which the text is rotated or skewed will output results that do not make sense. Additionally, there might be some photos where the text blends in with the background or is in a weird font. We attempted to fix these problems by pre-processing the image with a threshold to create more contrast between the text and background, but this give us results that were similar or worse than if we hadn't done any pre-processing. We also wanted to use PyEnchant's French dictionary to check whether the words that we get from text recognition are valid words, but we ran into the problem that some of the words we want might not be in the dictionary, the nonsense output ends up being a valid word, and most importantly, it seems like you need to install the french dictionary separately in order to use it.


ryaanahmed (changes requested):
This is looking really good.

@JasonLin43212 and I connected on Slack about this: he's going to take a shot at installing Tesseract in the GitHub Actions workflow, which'll be necessary to run the tests for this analysis module.

JasonLin43212:
@ryaanahmed I believe that I have added the installation of tesseract to the YML file. However, the backend test seems to be giving a really weird error: 
```
qt.qpa.xcb: could not connect to display 
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/cv2/qt/plugins" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
```

JasonLin43212:
@ryaanahmed I added the Tesseract to the YML file!

ryaanahmed (approved):

PR #71, opened 2020-10-16
Foreground detection

sharabhuiyan:
This analysis module extracts the foreground of an image and calculates its percentage in the photo. 

TO-DO:

- [ ] Increase the efficiency of the program. (Note: The program runs fairly quickly when we test individual photos but takes several minutes per photo when we test it on the collection.)

- [ ] Work on returning what the mask actually looks like so we can easily compare what the computer detects to be the foreground to the actual foreground. 

@ryaanahmed  @clarkce


ryaanahmed (approved):
This is great work as a start -- the code is very clean and clear.

I'll take a look myself at the performance issues on the whole dataset. I suspect that we'll want to scale the images down before actually running the grabcut analysis.

PR #65, opened 2020-10-09
Vanishing Point Analysis

taokinbo:
This PR uses Canny edge detection and Hough Line to detect vanishing points within the photo. 

Currently our analysis:

- [ ] Detects and extends significant (nonvertical and nonhorizontal) lines

- [ ] Returns the coordinate of the point that has the minimized sum of distances to all the lines detected

TODO

- [ ] Tune parameters of Hough lines and Canny to filter out insignificant lines and find more accurate lines

- [ ] Change algorithm to detect intersections rather than minimized distances between coordinates and lines

@ryaanahmed @clarkce

PR #74, opened 2020-10-20
Vanishing Point Analysis 

taokinbo:
This PR uses Canny edge detection and Hough Line to detect vanishing points within the photo.

Currently, our analysis:

- [ ] Detects and extends significant (nonvertical and nonhorizontal) lines

- [ ] Returns the coordinate of the average point of a group of coordinates with the most intersections or None if no coordinate was found

TODO

- [ ] Tune parameters of Hough lines and Canny to filter out insignificant lines and find more accurate lines

@ryaanahmed @clarkce


backend/app/analysis/find_vanishing_point.py
```diff
@@ -0,0 +1,220 @@
...
+            # only put line in filter_lines if the angle is NOT within epsilon of
+            # horizontal/vertical
+            epsilon = np.pi / 16
+            if (point_2_x - point_1_x) != 0 and abs(theta - np.pi / 2) > 2 * epsilon and theta > \
+                    epsilon:
+                # cv2.line(image, (point_1_x, point_1_y), (point_2_x, point_2_y), (0, 0, 255), 3, 8)
+                filter_lines.append(line)
+        except ZeroDivisionError:
+            pass
+    # van_point = find_van_coord(filter_lines, image.shape[0], image.shape[1])
```
ryaanahmed, line 52: Here's an example of what I'm talking about in the review message -- all of this is commented out and really should be deleted, as it makes the routine difficult to read.

ryaanahmed (changes requested):
The new algorithm looks like it gives better results! Great work there.

Since initial review over in #65, there's now a lot of unused code that's still here but just commented out.

Please either remove these commented out lines if they're actually no longer useful, or in the case of something like the original `find_van_coord` (where it's actually useful to have the original implementation hanging around, as we might want to repurpose parts of it), please comment why it's still there.

taokinbo:
@ryaanahmed 
I have removed all the unnecessary comments and changed our tester to fit the format of the other tester functions. There shouldn't be any merging issues now.

ryaanahmed:
This work rolled into #85, so closing this PR, and merging the work in over there.

PR #78, opened 2020-10-23
(Updated) Fix AnalysisView

JusticeV452:
- Changed AnalysisResult result field type back to TextField
- added parsed_result method to AnalysisResult model
- Updated the API call to sort photos by deserialized result values

I recopied all of the AnalysisView files to a branch based on master because the checks were not running.


ryaanahmed:
Thanks for the fixes! (comments were over at #64) The latest version was missing the URL configs, so I've restored those, and this is now good to go. Merging.

ryaanahmed (approved):

PR #79, opened 2020-10-23
Make syncdb treat Google Drive filenames as case insensitive

### LINKED ISSUE #63: Fix syncdb to make .jpg extension not case sensitive
- modify syncdb so that whether the extension is .jpg or .JPG doesn't matter

ryaanahmed:
This PR has syncdb lowercase filenames taken from the Google Drive photo repository before adding them to its internal lookup dict.

This means folks cleaning images no longer have to change uppercase `JPG` extensions to `jpg` by hand.

Fixes #63


ryaanahmed:
There were also a few 'jpeg' extensions in there, so we're robust to that now, too.

JusticeV452 (approved):

PR #84, opened 2020-11-06
Improved Photographer View 

JasonLin43212:
- Added a new flag to syncdb called `create_thumbnails` to `syncdb` which will create thumbnail versions of the cleaned versions of photos (but will use the front side if the cleaned side doesn't exist) and upload them into the Google Drive. Also added a `thumbnail_src` to the Photo model.
- Created a gallery view of the photos that a photographer has taken. Uses the thumbnails for faster loading times.
- Added analysis results for photographers based on aggregate analysis of the photos that they took.
- Run analysis now pickles the results at intervals so progress won't be loss from errors.


ryaanahmed (approved):
The new photographer view looks / works great. Particularly great to have the aggregate analyses -- that's very powerful. Some future work:

- page needs to be more accessible (menu bar item to a page of photographers?) -- right now can only get to it by clicking through from an individual photo

- maybe pushing the aggregate analysis stuff into the backend would allow us to create a view that sorts photographers by various analyses

merging.

ryaanahmed:
Also: syncdb with the `--create_thumbnails` flag doesn't actually set the `thumbnail_src` field, I think. We'll need to fix that at some point.

PR #85, opened 2020-11-20
Added visual analysis functionality and edited backend analyses

jlin00:
- added a widget that allows users to render analyses that are more visual in nature (i.e. drawing perspective lines, masking the background)
- edited the foreground_percentage and find_vanishing_point analyses to return visual information
- added resize functionality so that SVG is dynamic


backend/app/analysis/find_vanishing_point.py
```diff
@@ -0,0 +1,227 @@
...
+                and theta > epsilon
+            ):
+                filter_lines.append({
+                    '1_x': int(point_1_x),
+                    '1_y': int(point_1_y),
+                    '2_x': int(point_2_x),
+                    '2_y': int(point_2_y),
+                })
+
+        except ZeroDivisionError:
```
ryaanahmed, line 54: Since we can predict when this fires, my preference would be to if on that condition, and pass there. So, we'd have something like:

```
if point_2_x  == point_1_x:
    pass
```

earlier in the function. That way we don't have to wrap the whole thing in try/except blocks, and the edge case handling is clearer to the reader.

backend/app/analysis/find_vanishing_point.py
```diff
@@ -0,0 +1,227 @@
...
+def find_van_coord_intersections(lines):
+    """
+    Find the approximate vanishing point of the image by choosing the most representative point
+    from the largest cluster of intersections between lines in 'lines'
+    :param lines: list of lists containing start and end points of all filtered lines
+    :return: the average coordinate of the largest cluster of intersections
+    """
+    intersections = {}
+    tolerance = 30
+    # print(len(lines))
```
ryaanahmed, line 92: cleanup, plz remove

backend/app/analysis/find_vanishing_point.py
```diff
@@ -0,0 +1,227 @@
...
+    # return None if the image likely has many unnecessary lines (e.g. if there's a tree)
+    if len(lines) > 300 or len(lines) < 2:
```
ryaanahmed, line 94: please name all magic numbers. (300 is an arbitrary bound, so if you call it something like MAX_ALLOWED_LINES, then readers will know why it's there.)

backend/app/analysis/find_vanishing_point.py
```diff
@@ -0,0 +1,227 @@
...
+                for coord in intersections:
+                    distance = ((intX - coord[0]) ** 2 + (intY - coord[1]) ** 2) ** (1 / 2)
+                    if distance < tolerance:
+                        intersections[coord].append(intersection)
+                        found = True
+                if not found:
+                    intersections[intersection] = [intersection]
+    max_frequency = 0
+    van_point_list = []
+    # finds the largest cluster of intersections
```
ryaanahmed, line 115: nitpick -- I like to add a new line between chunks of code like this, for ease of reading.

backend/app/analysis/find_vanishing_point.py
```diff
@@ -0,0 +1,227 @@
...
+        return None
+    sum_point = [0, 0]
+    # finds the average of all coordinates in the largest cluster
+    for point in van_point_list:
+        sum_point[0] += point[0]
+        sum_point[1] += point[1]
+
+    sum_point[0] = round(sum_point[0] / max_frequency)
+    sum_point[1] = round(sum_point[1] / max_frequency)
+    # print(tuple(sum_point))
```
ryaanahmed, line 131: cleanup plz remove debug prints

backend/app/analysis/find_vanishing_point.py
```diff
@@ -0,0 +1,227 @@
...
+    # finds the largest cluster of intersections
+    for coord in intersections:
+        if len(intersections[coord]) > max_frequency:
+            max_frequency = len(intersections[coord])
+            van_point_list = intersections[coord]
+
+    if max_frequency == 0:
+        return None
+    sum_point = [0, 0]
+    # finds the average of all coordinates in the largest cluster
```
ryaanahmed, line 124: mean?

backend/app/analysis/foreground_percentage.py
```diff
@@ -36,8 +36,17 @@ def analyze_numpy_photo(photo):
     Returns percentage of a numpy photo that is part of the foreground
     """
     foreground_mask = grabcut_analysis(photo)
+
+    black_pixels = []
+    for i in range(0, len(foreground_mask), 20):
```
ryaanahmed, line 41: Not clear to me why we're limiting this range -- can you please comment and explain?

ryaanahmed: plus label any magic numbers...

frontend/src/photoView/photoView.js
```diff
@@ -11,11 +11,135 @@ const SIDES = {
     BINDER: 'binder',
 };
 
+export class FVDisplayWidget extends React.Component {
```
ryaanahmed, line 14: `FVDisplayWidget` -- didn't know what `FV` was when I first read this. Plz rename to something more explicit / not abbreviated so other readers will know.

frontend/src/photoView/photoView.js
```diff
@@ -11,11 +11,135 @@ const SIDES = {
...
+    render() {
+        const items = [];
```
ryaanahmed, line 16: please avoid generic names -- `lines`  or `line_arr` would be great here

ryaanahmed: whoops, not lines. it's all the svg paths, right?

frontend/src/photoView/photoView.js
```diff
@@ -11,11 +11,135 @@ const SIDES = {
...
+            items.push(<circle
+                cx={(this.props.vanishingPointCoord.x * this.props.width) / this.props.natWidth}
+                cy={(this.props.vanishingPointCoord.y * this.props.height) / this.props.natHeight}
+                r='10'
+            />);
+        }
+        return (
+            <div>
+                <svg
+                    className='analysis-overlay floatTL'
```
ryaanahmed, line 40: what's TL?

frontend/src/photoView/photoView.js
```diff
@@ -11,11 +11,135 @@ const SIDES = {
...
+                    width={this.props.width}
+                >
+                    {items}
+                </svg>
+            </div>
+        );
+    }
+}
+
+function configAnalysisFV(parsedValue, height, width, natHeight, natWidth) {
```
ryaanahmed, line 51: not gonna belabor this, bc we're talking this through in person. but you get the point.

frontend/src/photoView/photoView.js
```diff
@@ -11,11 +11,135 @@ const SIDES = {
...
+    width: PropTypes.number,
+    natHeight: PropTypes.number,
+    natWidth: PropTypes.number,
+};
+
+export class FPDisplayWidget extends React.Component {
+    render() {
+        const items = [];
+        const ratio = this.props.width / this.props.natWidth;
+        for (let i = 0; i < this.props.blackPixels.length; i++) {
```
ryaanahmed, line 81: you could use a `for pixel of this.props.blackPixels` here

frontend/src/photoView/photoView.js
```diff
@@ -11,11 +11,135 @@ const SIDES = {
...
+FPDisplayWidget.propTypes = {
+    percent: PropTypes.number,
+    blackPixels: PropTypes.array,
+    height: PropTypes.number,
+    width: PropTypes.number,
+    natHeight: PropTypes.number,
+    natWidth: PropTypes.number,
+};
+
+const VISUALANALYSES = {
```
ryaanahmed, line 130: VISUAL_ANALYSES plz

frontend/src/photoView/photoView.js
```diff
@@ -38,7 +166,14 @@ export class PhotoView extends React.Component {
             photoData: null,
             displaySide: '',
             availableSides: [],
+            view: 0,
+            width: 1,
```
ryaanahmed, line 170: I would probably initialize all of these values to `null` rather than `1`, bc `1` looks like real data, whereas you can check against `null` for whether these values were actually set at all. (Even though we don't do that here...)

frontend/src/photoView/photoView.scss
```diff
@@ -1,3 +1,23 @@
 .image-photo {
     width: 100%;
 }
+line {
```
ryaanahmed, line 4: d'oh, we don't have a css linter for this project. spaces after colons and consistent indentation, plz.

ryaanahmed (changes requested):
Really great work -- this brings these analyses to life. A few things for later improvement (maybe make follow-on issues):

- The location of the dropdown isn't visible (it's below the fold) on my laptop screen, so this very important piece of functionality would be easy to miss. Maybe we want to make the interface closer to the image?

- Maybe we want to start adding help text about what the drawn analyses actually are? The significant lines thing isn't totally clear, nor is the labelling of the vanishing point. Actually, VP isn't labelled at all.

REALLY GOOD. A lot of nitpicks within (as discussed in our call), please fix up and this is good to merge.

jlin00:
We have made the required changes. @ryaanahmed

ryaanahmed (approved):

ryaanahmed:
Thanks. Looks good. Merging!

PR #86, opened 2020-11-20
Created view to see the top 10 similar photos for a given photo

eebanerjee:
- Created All Photos View to show all images sorted by map square
- When a photo in All Photos View is clicked, it opens the Similarity View
- Similarity View displays the 10 most similar photos given the photo in the URL


montsegarza:
@ryaanahmed We fail python linters, but it's in the backend so we don't know if we should invest time on that bit. Let us know!

ryaanahmed:
> @ryaanahmed We fail python linters, but it's in the backend so we don't know if we should invest time on that bit. Let us know!

@montsegarza nope, don't worry about it -- that linter failure is in code that _I_ wrote, so I'll lint it myself!

ryaanahmed:
The work here has also been merged into #93, so I'm going to close out this PR, merge this branch into that to make sure we haven't missed anything, and then review in one swoop over on #93.

PR #88, opened 2020-11-24
Common Object Aggregator with YOLO Object Detection Model

sharabhuiyan:
- Added YOLO model to detect objects in photos.

- Used object detection to keep track of objects and their frequency in each photo.  This enables us to search up photos by object. 

**NOTE:** I could not push the YOLO weights to the repository because they are too large. Please download YOLOv3-608 from https://pjreddie.com/darknet/yolo/ and add it to the yolo files folder before running yolo model analysis. 

TO-DO:

- [ ] Attempt to integrate other object detection models into the project such as RetinaNet, RCNN, and Faster RCNN.

@ryaanahmed


ryaanahmed:
Thanks, all! The backend test is failing because the weights file is missing, but I'll add a step to the CI job that downloads the weights before running the tests.

The one failing linter rule is also probably okay -- I very often will disable `too-many-locals` if a single straight-line function seems like the best expression of the code in question. I'll take a look.

ryaanahmed:
Quick question -- this PR adds a few dependencies...
```
matplotlib~=3.3.2
gluoncv~=0.8.0
mxnet==1.6.0
```
... but I don't see them being used anywhere. Were these used in earlier experiments but not in the final code?

ryaanahmed (approved):
SUPER. Works great! A very significant contribution to the project. The code is super easy to read and very well documented.

I added some instructions to download the weights if they aren't there, so the analysis doesn't just crash with a cryptic file loading error from opencv. Also instructed the CI system to download and cache the weights file, so that the tests can run. See changes to .github/workflows/ci.yml if you're interested in how that works.

Little nitpick note: please be sure to remove any cruft that may have accumulated during development when making a PR. See my commits where I removed some files that were unnecessarily duplicated, a constants file that had become redundant at some point along the way (coco_names.py), and some extra dependencies that weren't being used anymore. This kind of stuff can slowly accrete in a project and make the thing messy over time. A good way to avoid is just to take quick scan through the "Files changed" right when you make the PR, so you can see if anything seems to be out of place, and to delete files you no longer need while you're working as soon as you find they're not useful. Remember that since you have a commit history, you can always rollback if you realize you got rid of something you need!

That nitpick paragraph was disproportionately long. This is fantastic work. Merging!

PR #89, opened 2020-11-28
Created buttons to navigate from photo to photo

taokinbo:
- [ ] Created buttons that link the photos to each other without having to head back to homepage
- [ ] Each photo is linked to the previous photo and the next photo 
- [ ] Buttons are placed at top of page and at bottom of page
- [ ] Photos follow the order of the homepage index
- [ ] When button is hovered over, the color turns black
- [ ] Last photo in list loops back to the first photo

NextButton
NextButtonHover

@ryaanahmed @clarkce


ryaanahmed:
Looks good, works well. Going to go ahead and merge.

A few of things to think about for next steps:
- This now pulls in a _lot_ of extra data since it's calling the all map squares API endpoint. We're not using most of it for this page, so it's probably worth constructing an API endpoint that sends less stuff for this functionality, for speed.
- Would be great to have some information about next and previous photos (thumbnail, photo number, map square, etc.)
- The state when there's no previous photo isn't very clear

ryaanahmed (approved):

PR #91, opened 2020-12-07
Kmeans image clustering

JusticeV452:
Use k-means model to group similar images
- Added Cluster model and serializer to the database to manage photo groupings
- Added custom django command to fit a k-means model to the photo database and create cluster instances
- Added frontend view to view photo groupings
- create clusters by running "python manage.py createkmeans [number_of_clusters]"


ryaanahmed (approved):
Super cool. Would be interesting to see whether we get different/better/worse results by using feature vectors extracted in #93 to do this clustering as follow on work.

I made a light refactor to the pickling code; there was a branch where it would just load and then resave, and I think that no longer happens.

Probably the management command wants some factoring out into individual functions in order to extend this work, but that can happen as a next step. Also would love a frontend view that shows all of the clusters in on a single page in some kind of interactive format.

PR #92, opened 2020-12-09
Interactive Map View

L-Jewel:
- Replaces landing page with an interactive map
- Map displays the map squares in regularly shaped map page grids
- Upon clicking the map squares, a pop up containing the map square number, the number of photos in the map square, and a link to its map square page appears.
- If there are no photos in a map square, the square is grayed out, while the ones that contain photos appear in red


joshfeli:
We resolved all basic linting errors, but we would have to refactor the `populate_database` function in `syncdb` to pass the python linter test.

ryaanahmed:
> We resolved all basic linting errors, but we would have to refactor the populate_database function in syncdb to pass the python linter test.

Thanks -- I added a comment to the function explaining what's going on. `populate_database` is at a stage where it does want a cleanup refactor, but let's not do that in this PR, both for expediency and so that the PR is intelligible. At this point we need to separate `populate_database` into a separate function per model, as the models have diverged significantly over the course of the semester.

ryaanahmed:
Just posting a screenshot here...

map_view

ryaanahmed (approved):
This is great work -- the result really brings the project to life!

Please take a look at some of the changes I made, particularly in my commit b5897bc to the populate_map_pages module. I'd usually do this as review, but given how close we are to wrapping up, went ahead and made the changes myself. Some things to note:

- `csv.reader` and `csv.writer` are designed to work with lists and such -- take a look at the final version here for an example. You don't want the `delimiter=' '` -- by default the delimiter is a comma, so you don't need to split on commas, etc. The reader will just spit out rows already split up by comma.

- In general, you don't want to do all of the work while you have file handles open. Better to open a file, grab whatever data you need from it, do the computation and cache the results to some kind of data structure in memory, and at the _end_ open another file handle to write out the results. Again, see `populate_map_pages` after my last couple of commits for an example.

Great work, though. Merging!

ryaanahmed:
One more note -- this PR changes a few heights etc. to be defined in `vh`. You have to be a bit careful with setting things according to viewport height because it's very easy to end up in situations where, e.g., the navbar is going to overlap other UI elements.

I'm going to do a pass to cleanup CSS problems at the end after merging everything, but just FYI many of the changes to `vh` and `vw` in this PR probably won't come out the other end of that because at certain viewport dimensions, things look screwy.

e.g. you can end up with situations like this:

vh_problems

PR #93, opened 2020-12-10
Add backend methods for determining photo similarity

Dylan-Weber:
- Added pairwise similarity and mean squares error analyses.
  - For the most part, these give very similar results to cosine similarity
- Added maps between different layers of resnet18 and their required vector sizes so that different layers can be experimented with. Vicky said maxpool and Pairwise gave the best results. 
  - Once again, changing the layer did not seem to change the results too much.

Unfortunately, I don't think this project turned out to be very successful. The results for each of the photos often tend to give the same similar photos, so the results are not indicative of actual similarity. Still, I hope this can be used as a potential base for anyone wishing to measure similarity as part of Catherine's UROP team.


ryaanahmed:
Thanks for great work on this project. It's a bummer that we weren't able to get better results, but the infrastructure here will absolutely be essential for follow-on work on similarity that the next folks working with @clarkce on this project will undertake.

My suspicion is that to get better results, we really need to train our own model, since the stuff in imagenet is so different from our images, and in order to train our own model, we'll need more images, which is going to take a little while with the scanning at the archives. Once we're able to do that, almost all of the code will be able to be directly repurposed.

The organization of the different similarity functions, the frontend, etc. are all great.

Thanks to all who worked on this project for being willing to try something new and difficult!

ryaanahmed (approved):

PR #103, opened 2021-01-13
Improved Loading Speed and Loading Page

JasonLin43212:
- Improved the loading speed of the landing page (~5s -> ~3s)
- Improved loading speed of Photo page (~4s -> almost instant)
- Added a loading screen:

Screenshot from 2021-01-13 14-36-59


frontend/src/clusterView/clusterView.js
```diff
@@ -54,9 +54,13 @@ export class ClusterView extends React.Component {
 
     render() {
         if (this.state.loading) {
-            return (<h1>
-                Loading!
-            </h1>);
+            return (
+                <React.Fragment>
```
ryaanahmed, line 58: This pattern is in a bunch of places as a loading screen component -- let's factor that out.

ryaanahmed (changes requested):
Looks really good - let's just factor out the whole loading screen, since that same fragment appears in every view now.

ryaanahmed:
p.s. the speed changes are _really_ helpful.

PR #104, opened 2021-01-15
Changed the Map and Square Colors

adanna-a:
- Picked a different open street map layer with less roads and colors
- Change empty squares to darker grey 
- Changed squares with pictures from red to blue

Screen Shot 2021-01-15 at 10 53 19 AM


frontend/src/index/index.js
```diff
@@ -67,7 +74,10 @@ class Map extends React.Component {
                                     <Rectangle
                                         key={index}
                                         bounds={mapSquareBounds}
-                                        color={'#b3b3b3'} weight={2}
+                                        color={'#737a7a'} weight={1}
```
ryaanahmed, line 77: Rather than hard coding color here, can you instead give this `Rectangle` and the one below CSS classes, and in the corresponding scss file point to a color variable? @L-Jewel is working on overall color scheme, so then she can tweak in the scss files without having to go hunting for these.

p.s. the curly braces in JSX mean "this is a JavaScript expression" so for strings you don't need them: `color='#737a7a'` resolves to the same as `color={'#737a7a'}`. You need the curly braces if instead of a string or number literal, you were passing a more complex expression like a function call or doing some addition.

ryaanahmed (changes requested):
Looks a lot nicer! Let's factor out the color stuff (see review comments) to make @L-Jewel's color scheme work easier, and then it's good to go.

frontend/src/index/index.js
sharabhuiyan: @ryaanahmed I have assigned both rectangles CSS classes.

ryaanahmed (approved):
thanks for the changes - looks great. merging!

PR #107, opened 2021-01-19
Search Page, About Page, Navbar Update, and Analysis Page

JasonLin43212:
Note: Padding and margins for the top and bottom of the page are a bit weird. This problem is fixed on the image-page pull request.
Another Note: I'm not sure why but there are a lot of files in node modules that were modified, which is why it says that there are over 2000 files that were changed

In this branch, we were able to:
- Implement the search functionality:
    - You can search by a full text search, which will use the text you input and check if photographer names, number, photo tags, 
       or photo caption match with the text.
    - You can also do an advanced search, which will allow you to search for specific photographers, captions, or tags individually.
Screenshot from 2021-01-19 12-01-47

- Updated the navbar for easy navigation to different pages.
Screenshot from 2021-01-19 12-03-44

- Updated the about page with lorem ipsum text along with the photos of everyone who worked on the project.
- Added an all analysis page for easy navigation to the analysis pages. A link will only show up if there is at least one result for a specific type of analysis.
Screenshot from 2021-01-19 12-07-05
- Map Square page now uses photos
Screenshot from 2021-01-19 12-08-41
- Fixed bug in syncdb where it incorrectly assigns the photographers for photos with no photographer number

(Sorry about the gigantic pull request)

Some things that we should examine after all the pull requests are merged:
- Navbar and Footer padding on all the pages
- Making sure that the map on the landing page is the right size and has visible buttons for zooming in and out.
- Dropdown for tag selection on search page is not covered by footer


ryaanahmed:
> Another Note: I'm not sure why but there are a lot of files in node modules that were modified, which is why it says that there are over 2000 files that were changed

@JasonLin43212 it looks like someone inadvertently ran `npm install` from the project root directory rather than in the `frontend` directory, which is why it created a new `node_modules` dir in the root we've got all the extra files. Please delete this new project root level `node_modules` dir.

ryaanahmed:
(`frontend/node_modules` is git ignored, so changes in node_modules shouldn't show up in the repo at all)

JasonLin43212:
@ryaanahmed I just deleted the `node_modules` directory.

backend/app/views.py
```diff
@@ -188,3 +203,63 @@ def get_photos_by_cluster(request, number_of_clusters, cluster_number):
...
+    is_advanced = query['isAdvanced']
+
+    if is_advanced:
+        photographer_name = query['photographerName']
+        photographer_num = query['photographerId']
+        caption = query['caption'].strip()
+        tags = query['tags']
+
+        django_query = Q()
+        if photographer_name != '':
```
ryaanahmed, line 223: not preventing merge but -- empty strings and lists of length 0 are "falsy" in Python, so these can just be `if photographer_name`, `if tags:` etc.

backend/app/views.py
```diff
@@ -188,3 +203,63 @@ def get_photos_by_cluster(request, number_of_clusters, cluster_number):
     cluster = Cluster.objects.get(model_n=number_of_clusters, label=cluster_number)
     serializer = PhotoSerializer(cluster.photos.all(), many=True)
     return Response(serializer.data)
+
+
+@api_view(['POST'])
+def search(request):
```
ryaanahmed, line 209: Great work putting this search query together.

frontend/src/about/about.js
```diff
@@ -103,7 +109,42 @@ class About extends React.Component {
                     </div>
                     <div>
                         <p>
-                            Copy about the project will go here.
+                            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
```
ryaanahmed, line 112: Please open a follow-up issue to make sure we don't forget to fill this in with real text soon!

frontend/src/search/search.js
```diff
@@ -0,0 +1,315 @@
...
+        this.setState({ ...searchData });
+    }
+
+    async componentDidMount() {
+        try {
+            const searchTagResponse = await fetch('/api/get_tags/');
+            const searchTags = await searchTagResponse.json();
+            const { tags, photographers } = searchTags;
+            // Sort by name and then by number, if the photographers have one
+            photographers.sort((a, b) => {
```
ryaanahmed, line 224: Cleanup TODO (not preventing merge) -- looks to me like this sort can be done in the backend in the db query by Django using the `order_by` method of a queryset (https://docs.djangoproject.com/en/3.1/ref/models/querysets/#order-by). Much faster to do a sort this way if it's possible than doing it in the frontend like this.

frontend/src/search/search.js
```diff
@@ -0,0 +1,315 @@
...
+                            this.state.data
+                            && <div>
+                                <h2>
+                                    {this.state.isAdvanced
+                                        ? 'Advanced Search Results'
+                                        : 'Search Results'}
+                                </h2>
+                                <p>{this.state.searchText}</p>
+                                <div className='search-results'>
+                                    {this.state.data.map((photo, k) => {
```
ryaanahmed, line 281: Cleanup TODO (not preventing merge) -- I'd like to see this factored out into a component for readability.

ryaanahmed:
> @ryaanahmed I just deleted the `node_modules` directory.

Great! Note that there was also an errant `package-lock.json` file that was created at the same time. I've gone ahead and deleted that.

ryaanahmed (approved):
Great work. Some comments within for future cleanup, but merging now.

PR #111, opened 2021-04-03
generates heat map to show photo density in each square

sophzheng:
The map on the home page is now colored in based on a gradient of 5 colors (buckets). These buckets represent the density of photos in the square, calculated as a percentage of the number of photos in the square with the maximum number out of all of the squares. Currently, the colors go from pink (0%) to purple (100% of the max). 

Because the bucket sizes (photos per bucket) are calculated as a percentage of the max photos, they should not need to be altered in the future as more photos are added to the database. 

There is also a legend to help the user navigate which colors (buckets) correspond which range of number of photos for the square, as seen in the top right. 

Screen Shot 2021-04-23 at 1 39 27 PM

### DIFF
--- a/frontend/src/index/index.scss
+++ b/frontend/src/index/index.scss
@@ -76,38 +76,32 @@ code {
     stroke-opacity: 0.5;
 }
 
-//.map-square-box {
-//    stroke: steelblue;
-//    stroke-width: 2.8;
-//    fill: dodgerblue;
-//    fill-opacity: 0.3;
-//}
 .map-square {
     stroke-width: 2.8;
     fill-opacity: 0.5;
 }
 
-.map-square-box-one {
+.box-one {
     stroke: #E85285;
     fill: #E85285;
 }
 
-.map-square-box-two {
+.box-two {
     stroke: #C9458B;
     fill: #C9458B;
 }
 
-.map-square-box-three {
+.box-three {
     stroke: #A93790;
     fill: #A93790;
 }
 
-.map-square-box-four {
+.box-four {
     stroke: #8A2995;
     fill: #8A2995;
 }
 
-.map-square-box-five {
+.box-five {
     stroke: #6A1B9A;
     fill: #6A1B9A;
 }
--- a/frontend/src/index/index.scss
+++ b/frontend/src/index/index.scss
@@ -76,9 +76,38 @@ code {
     stroke-opacity: 0.5;
 }
 
-.map-square-box {
-    stroke: steelblue;
+//.map-square-box {
+//    stroke: steelblue;
+//    stroke-width: 2.8;
+//    fill: dodgerblue;
+//    fill-opacity: 0.3;
+//}
+.map-square {
     stroke-width: 2.8;
-    fill: dodgerblue;
-    fill-opacity: 0.3;
+    fill-opacity: 0.5;
+}
+
+.map-square-box-one {
+    stroke: #E85285;
+    fill: #E85285;
+}
+
+.map-square-box-two {
+    stroke: #C9458B;
+    fill: #C9458B;
+}
+
+.map-square-box-three {
+    stroke: #A93790;
+    fill: #A93790;
+}
+
+.map-square-box-four {
+    stroke: #8A2995;
+    fill: #8A2995;
+}
+
+.map-square-box-five {
+    stroke: #6A1B9A;
+    fill: #6A1B9A;
 }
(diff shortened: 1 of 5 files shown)


frontend/src/index/index.scss
```diff
@@ -76,9 +76,38 @@ code {
...
+//    stroke-width: 2.8;
+//    fill: dodgerblue;
+//    fill-opacity: 0.3;
+//}
+.map-square {
     stroke-width: 2.8;
-    fill: dodgerblue;
-    fill-opacity: 0.3;
+    fill-opacity: 0.5;
+}
```
MBJean, line 88: It looks like this `.map-square` class isn't getting applied anywhere. It turns out that the default styling supplied to the `<Rectangle>` element by our external library almost perfectly matches the styling supplied by this class. You can investigate that through the `Elements` tab of the Chrome Dev Tools. The consequence is that it's not visually apparent that this class isn't actually doing anything! We can fix resolve this by adding it to the string we assign in `mapSquareBucket`.

Screen Shot 2021-04-06 at 4 01 55 PM

MBJean (changes requested):
This is a great first PR to contribute to the project! I've left a few comments below with a few things to address. Please review my feedback and let me know if you have any questions. With all of my reviews, please always feel comfortable pushing back or explaining your reasoning. If you have any questions or needs clarification, don't hesitate to respond via the comments in this PR or reach out and we can meet in office hours, the Friday hack session, or schedule some specific time to meet.

sophzheng:
@MBJean Hi, Mike! Just updated the scss classes to be more specific and pushed the changes. Thank you!

MBJean:
@sophzheng @NinaLiYH thanks for addressing my feedback! Your changes look good to me. You're correct that that `python_linters` failures has nothing to do with your changes. The tldr; for what happened is that the Python package that we use in the `python_linters` GitHub action, `pylint`, upgraded itself from version `2.7.4` to version `2.8.1` in your most recent commit. Version `>=2.8` includes a new linter check `consider-using-min-max-builtin` (which you can see in the `pylint` changelogs [here](https://github.com/PyCQA/pylint/blob/master/ChangeLog#L78)). Since `pylint` checks the whole codebase each time it runs, not just the changes in this PR, and because it's now checking the codebase at a new version with additional checks, it flagged something that it hadn't previously.

If you're interested in how I came to that conclusion, here's a more detailed version of what I was looking at:
- As you did, I checked the [logs](https://github.com/dhmit/paris_1970/pull/111/checks?check_run_id=2432519095) for the `python_linters` GitHub action to see what failed, and saw that the linter failed something that you didn't change in this PR.
- My first assumption was that something in our various downloaded packages was now being downloaded at a different version than it was in past runs of those jobs. Failed CI/linters/compilation/etc. jobs (i.e., things that run in environments other than your local machine) are usually a good sign that some package has had its version changed. You can look through the full logs of those GitHub action to see that we download all of our Python packages when we run those actions, so that's an opportunity for some unintended/unexpected change to creep in.
- With that assumption in mind, I looked through the GitHub action logs for `python_linters` for both your [second-to-last commit](https://github.com/dhmit/paris_1970/runs/2421927251) and your [most-recent commit](https://github.com/dhmit/paris_1970/pull/111/checks?check_run_id=2432519095). I could see that in the `Install Python packages` job, your previous commit was downloading `pylint` at version `2.7.4` ([link](https://github.com/dhmit/paris_1970/pull/111/checks?check_run_id=2432519095#step:4:159)), while your most recent commit was downloading that package at version `2.8.1` ([link](https://github.com/dhmit/paris_1970/pull/111/checks?check_run_id=2432519095#step:4:159)).
- Knowing that there was a discrepancy between those versions, I visited the `pylint` [GitHub page](https://github.com/PyCQA/pylint) to see if they have a changelog, which they happily do ([link](https://github.com/PyCQA/pylint/blob/master/ChangeLog)), and further to check if something might have changed between those two versions that would explain the new failure. I skimmed down to the section entitled `What's New in Pylint 2.8.0?` and found [this line](https://github.com/PyCQA/pylint/blob/master/ChangeLog#L78), which explains that a new check was added to `pylint` that does indeed explain the new failure.
- Finally, it's worth thinking about _why_ `pylint` updated itself in your most recent commit. Generally, we try to keep packages fixed either at specific versions or at least to minor versions so we don't get unexpected changes. You can see in our [`requirements.txt`](https://github.com/dhmit/paris_1970/blob/master/requirements.txt), for instance, that we 'pin' most of our packages to specific versions using the `==` syntax (full syntax explanation in the [docs](https://iscompatible.readthedocs.io/en/latest/)). In this case, we install the package `pylint` not in our `requirements.txt`, but rather in the scripts that tell GitHub how properly to run the linters ([here](https://github.com/dhmit/paris_1970/blob/master/.github/workflows/ci.yml#L54)), where we do _not_ indicate that we want to pin our package download to a specific version. We can see from the ChangeLog linked above that `pylint` had a version update on `2021-04-24`, and so the very next commit to run that `pip install pylint` without a version pin brought in the new version.

MBJean (approved):
LGTM! Please see my (overly wordy) explanation above for why you saw that unexpected linter failure. I've pushed up a commit resolving it. This PR is now ready to merge into the main branch. Great work!

PR #112, opened 2021-05-18
Improving Search and Navigation

JusticeV452:
- Added ability to search for photos with a result for specified analyses. Analyses with numerical values (currently does not support categorical values) can also be filtered by a specified value range.
- Updated similar photos list to be confined to a scrollable box of a fixed height. Photos in the list are now in order of descending similarity.
- Added link to a photo's similarity page from the photo view page.
- Fixed SimlarityView to show the most similar photos instead of the least similar photos.
- SimilarityView url now includes the number of similar photos to view.


backend/app/views.py
```diff
@@ -26,9 +28,31 @@
...
+    'resnet18_mean_squares_similarity': 'photo_similarity.resnet18_mean_squares_similarity',
+    'resnet18_pairwise_similarity': 'photo_similarity.resnet18_pairwise_similarity',
+    'pop_density_detection': 'pop_density_detection',
+    'portrait_detection': 'portrait_detection',
+    'stdev': 'stdev',
+    'text_ocr': 'text_ocr',
+    'whitespace_percentage': 'whitespace_percentage',
+    'yolo_model': 'yolo_model'
+}
+
```
MBJean, line 55: This constant gave me a bit of pause, since I can see that it's returned in the response to the frontend, and looks to be a straightforward mapping of 'sanitized' analysis names to their module counterparts. Before I started digging into the frontend implementation details, my major concern was that it seemed like the frontend would somehow be handling backend concerns. Looking at how the `analysisTags` property on the response JSON is dissected and used throughout the `Search` and `SearchForm` components, I can see a ton of business logic. 

So, in short: just seeing this made me concerned that the frontend was going to be doing too much work, and though I've only had a few minutes to check out the implementation, I can see that it is indeed doing as lot of work. Would you be able to explain the goal of passing this map to the frontend?

frontend/src/similarityView/similarityView.scss
```diff
@@ -0,0 +1,13 @@
...
+
+.display-box {
+    display: inline-block;
+    flex-direction: row;
+    justify-content: space-between;
+}
+
+.analysis-page {
+    padding: $header-height 25px $footer-height;
+}
```
MBJean, line 13: One major point, two minor points:

Major:
This file isn't actually being compiled. Our current setup requires you to manually add new .scss files in `frontend/src/UILibrary/style.scss`.

Minor:
1. For the `.display-box` selector, without `display: flex`, I don't think that the `flex-direction` and `justify-content` properties are doing anything.
2. Because of how our SCSS is compiled, these are all 'global' selectors. In other words, the `display-box` class is being applied to across the application, so you can for instance see that multiple of those selectors are being applied to this one DOM element, because we select against `display-box` as a class in multiple `.scss` files. You may want to rename the class to be specific to this page view.

backend/app/views.py
JusticeV452: I think the only reason I passed the map to the frontend was to convert between the module name and the display name. However, the conversion is only needed when sending the search request to the backend, so I'll only pass the tags to the frontend and map them on the backend.

frontend/src/similarityView/similarityView.scss
JusticeV452: I've just removed `similarityView.scss` since the SimilarityView page is intended to use the same styling as the other pages and removed the `flex-direction` and `justify-content properties` from `analysisView.scss`. I also removed `.display-box` from `clusterView.scss` since it is also supposed to use the same style for `display-box`.

backend/app/views.py
MBJean: 👍

PR #113, opened 2021-12-07
Collective Photo Analysis and Population Density

JusticeV452:
- Added new population density calculation based on the output of the yolo_model
- Work-in-progress: Created 'statistics_analysis' function to calculate overall statistics for groups of photos in the database based on the result of other analyses.

### DIFF
--- a/backend/app/analysis/aggregated_analyses.py
+++ b/backend/app/analysis/aggregated_analyses.py
@@ -4,8 +4,8 @@ Module to aggregate analysis results for photos
 
 import os
 
-from functools import reduce
 from app.models import PhotoAnalysisResult
+from app.analysis.yolo_pop_density import object_density
 
 from django.conf import settings
 
@@ -128,11 +128,4 @@ def max_object_density(object_name):
     def run(yolo_dicts):
         densities = [object_density(object_name, yolo_dict) for yolo_dict in yolo_dicts]
         return max(densities)
-
     return run
-
-
-def person_density_calc(photo):
-    return object_density("person", PhotoAnalysisResult.objects.filter(
-        name="yolo_model", photo=photo
-    ).first().parsed_result())
--- a/backend/app/analysis/aggregated_analyses.py
+++ b/backend/app/analysis/aggregated_analyses.py
@@ -10,13 +10,12 @@ from app.analysis.yolo_pop_density import object_density
 from django.conf import settings
 
 CLASS_NAMES_DIR = os.path.join(settings.YOLO_DIR, 'coco.names')
-with open(CLASS_NAMES_DIR, "r") as file:
+with open(CLASS_NAMES_DIR, encoding='utf-8') as file:
     CLASS_NAMES = [line.strip() for line in file.readlines()]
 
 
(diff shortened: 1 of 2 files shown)


backend/app/analysis/aggregated_analyses.py
```diff
@@ -0,0 +1,131 @@
+"""
```
ryaanahmed, line 1: Since this module is WIP and not wired into anything, I'm leaving this alone for now: we can finish up in follow-up work.

backend/app/analysis/yolo_pop_density.py
```diff
@@ -0,0 +1,108 @@
...
+    Given an object name, the output of the yolo_model analysis for a photo (yolo_dict),
+    and the dimensions of the photo, return the density of the specified object
+    in the photo.
+    :param object_name: string (name of the object)
+    :param yolo_dict: dictionary (output of the yolo_model for the photo)
+    :param photo_dim: tuple (dimensions of the photo the object is located in)
+    :return: A number representing how densely the specified object type is packed
+    in an image. More overlapping objects corresponds to higher density.
+    """
+    boxes = *filter(lambda box: box['label'] == object_name, yolo_dict.get("boxes", {})),
```
ryaanahmed, line 78: Although this line works just fine, holy moly @JusticeV452, you might win an award here for writing the single most inscrutable but also ingenious line of Python I've ever read. Come find me in the lab sometime for a prize. I'm serious.

Here are a couple of slightly easier-to-read alternatives:

```
yolo_boxes = yolo_dict.get("boxes", {})
boxes = list(filter(lambda box: box['label'] == object_name, yolo_boxes))
```

Or (I think I prefer this):

```
yolo_boxes = yolo_dict.get("boxes", {})
boxes = [box for box in yolo_boxes if box['label'] == object_name]
```

We end up with a list rather than a tuple, so we don't have the protection of immutability, but for the use here, this seems fine to me.

backend/app/analysis/yolo_pop_density.py
```diff
@@ -0,0 +1,108 @@
...
+        density += density_i / (len(boxes) - 1)
+    return density
+
+
+def analyze(photo: Photo):
+    yolo_dict = PhotoAnalysisResult.objects.filter(
+        name="yolo_model", photo=photo
+    ).first().parsed_result()
+    image = photo.get_image_data()
+    photo_dim = image.shape[:2][::-1] if image else (10000, 10000)
```
ryaanahmed, line 107: `if image` fails here under some conditions with this error from numpy:
```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

Basically: the numpy notion here is that whether an array is truthy or not is ambiguous, as a user of the library might reasonably assume that an array is truthy if _all_ of its elements are truthy or if any elements exist (as in a Python list). What we probably want here is actually to if on the size, or `if image is not None`.

ryaanahmed (approved):
Looks good! I'm going to go ahead and fix up a couple of things myself. Will merge without fixing all the linter problems, as a couple of them are docstrings we want to make sure to add, but okay not to get to before we show this work this afternoon.

PR #166, opened 2022-06-15
Fix analyses on photos and refactored photo display code

JusticeV452:
- Updated `get_image_data` method for photo model to use local photo path when running an analysis
- Created a base `PhotoViewer` component with commonly used code for displaying photos implemented as a class method
- Pages photos are displayed on now extend `PhotoViewer` instead of `React.Component`.


backend/app/views.py
```diff
@@ -123,12 +123,15 @@ def all_map_squares(request):
...
 
 @api_view(['GET'])
-def get_photographer(request, photographer_number):
+def get_photographer(request, photographer_number=None):
     """
     API endpoint to get a photographer based on the photographer_id
     """
-    photographer_obj = Photographer.objects.get(number=photographer_number)
-    serializer = PhotographerSerializer(photographer_obj)
+    if photographer_number:
```
anastasia, line 130: good update, thanks!

frontend/components/PhotoViewer.js
```diff
@@ -0,0 +1,81 @@
...
+    return result;
+}
+
+export class PhotoViewer extends React.Component {
+    constructor(props) {
+        super(props);
+    }
+
+    getSource (photoData, displaySide = false) {
+        photoData.propTypes = {
```
anastasia, line 24: Thanks for moving this here. We will want to really simplify this function once we start using the new photos — No need for SIDES obj any more, either. Each photo will come with a slide view and a photo view and that's it.

anastasia (approved):
Great work!

PR #169, opened 2022-06-27
New photo display format, photo slider added, everything right font

nfrieden:
To make photos show up in your Map Square View (and in the Photo View itself) you will have to change the names of the photo paths in assets/images/photos/*map square number*, as shown in the example below: 
image
Photos will not show up in any folder unless you change their names to #_photo or #_slide (we are NOT using “cleaned” anymore). Don’t worry about there being extra blank squares on the map square view-- these won’t show up once we reflect file name changes in the database.

This pull request adds a horizontal photo slider to view thumbnails of similar photos. As part of this change, the photo serializer (in the serializer.py file) was changed to return parsed json analysis results rather than parsing the json on the frontend. The photo serializer now parses results from the similarity analysis into dictionaries with the similarity result representing photos. To view the photos in the slider, populate the database with similarity results by running these commands in the terminal:
python manage.py runanalysis photo_similarity.resnet18_feature_vectors
python manage.py runanalysis photo_similarity.resnet18_cosine_similarity

With correctly named photos and the similarity analysis run, PhotoView should look like this:
image


frontend/scss/home.scss
```diff
@@ -2,9 +2,7 @@ body {
     margin: 0;
     padding: 0;
     overflow-x: hidden;
-    font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen",
-    "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue",
-    sans-serif;
+    font-family: "eina_regular";
```
anastasia, line 5: You defined `$font-family-sans-serif` scss variable, now you can use it here!
`font-family: $font-family-sans-serif`

frontend/scss/home.scss
```diff
@@ -57,6 +55,10 @@ code {
     color: black;
 }
 
+.map-popup {
+    font-family: "eina_regular";
+}
```
anastasia, line 60: font-family: $font-family-sans-serif;

frontend/scss/photoView.scss
```diff
@@ -94,12 +94,88 @@ rect {
...
+.side-button {
+    font-size: 18px;
+    font-weight: 500;
+    border: 2px solid;
+    padding: 2px 10px;
+    background-color: white;
+}
+
+.side-button:hover {
+    background-color: #20CCD7;
```
anastasia, line 106: In a future PR, let's define our hex values inside of a variables file, like here: https://github.com/dhmit/self-sustaining-cities/blob/main/frontend/scss/variables.scss
Then we can use these variables everywhere. The reason for this is if we have to tweak any values, we would just have to change them once in that file as opposed to everywhere they're mentioned

frontend/scss/photoView.scss
```diff
@@ -94,12 +94,88 @@ rect {
...
+}
+
+.tag-button {
+    font-size: 16px;
```
anastasia, line 110: Let's use em values wherever possible

frontend/scss/photoView.scss
```diff
@@ -94,12 +94,88 @@ rect {
...
 
-.image-box{
+.image-box {
     position: relative;
+    padding: 25px;
+    background-color: black;
+}
+
+.similar-photo {
+    max-width: 100px;
```
anastasia, line 155: Since we want these to be responsive, let's use em values that will change depending on screen size

anastasia (changes requested):
Hey @nfrieden, I added some fix requests to the scss code here. When these are done, the PR is ready to merge.

anastasia:
Go team! 🚀

PR #170, opened 2022-06-28
New file for slide image cropping

czheng10:
Python script for cropping and removing the white background from photo slides. 

Note: Set to only work for white backgrounds

Takes in 3 parameters:
  1. old_path: (str) path to directory containing images 
  2. new_path: (str) path where output is intended to be saved in
  3. fuzz_val: (str) optional third parameter for the percentage value for Image Magick fuzz command, defaults to 20%


anastasia (commented):
A couple small change requests but good to go after!

anastasia:
Thanks for making those changes!

PR #171, opened 2022-06-29
New ParisMap Component with Layers

JusticeV452:
- Added new `ParisMap` component for use across the website
- Added new arrondissements.geojson file to backend/data
- Geojson data is served at /api/arrondissements_geojson/
- Geojson data for a single arrondissement is available at /api/arrondissements_geojson/[INSERT ARRONDISSEMENT NUMBER]

To add layers to the map, a list or dictionary (where keys are the name of the layer and the values are components to render) can be passed as the layers prop. The map can be configured to show multiple layers at a time or a single layer at a time using the `singleLayer` prop (bool). Finally, the layer selection window can be disabled by setting `layerSelectVisible` to false.

image
image
image


anastasia (changes requested):
Hi @JusticeV452, great work! Just some small cosmetic changes requested. Thank you.

PR #172, opened 2022-07-01
Blog functionality

Nisha-Nathan:
~~I created a template for the blog home page which is automatically populated with blogs whenever blog entries are done in the admin panel. I introduced two tables (Blog & Category) to keep blog info and enable multiple tags assignment to a single blog instead of being limited to a single category choice and may further be used to display blogs sharing the same tags. I am yet to integrate the cms blog home page with this template and acquire information from the created models.
I am yet to also customize the navbar as required.~~

This PR:
- adds blog app
- adds [BlogPost](https://github.com/dhmit/paris_1970/pull/172/files#diff-0aeb1a60bc89497816b8f2759a5b23b5425f571261c66499e13dffd2a4b18fa7) model which relies on the built-in User model for authorship
- Adds TinyMCE (WYSIWYG editor) for rich HTML content field on BlogPost model
- adds Tags from taggit which are not currently connected to BlogPost (TODO) 
- moves navbar to the [base.html file](https://github.com/dhmit/paris_1970/pull/172/files#diff-d448116430234e6d313e9295459f70cc2124cf7530d5f14039e023f2ae44668a)
- adds a hierarchy for `backend/templates` (index.html is built on top of base.html, added includes folder)
- Adds blog templates
- Removes some cruft from our app like the instructions popup modal and some console logs

### Note from @anastasia:
This PR requires `pip install -r requirements.txt` and `python manage.py migrate`!


requirements.txt
```diff
@@ -4,21 +4,42 @@ backcall==0.2.0
 cachetools==4.1.1
```
anastasia, line 4: Remember to add the requirements you're relying on after installing them! `pip install --freeze > requirements.txt`

backend/config/urls.py
```diff
@@ -72,3 +75,10 @@ def react_view_path(route, component_name):
     path('all_analysis/', views.all_analysis_view),
     path('clustering/<int:num_of_clusters>/<int:cluster_num>/', views.cluster_view),
 ]
+
+cms_urlpatterns = [
+    re_path(r'^blog/', include('cms.urls')),
+    path('cms/', cms_views.index),
```
anastasia, line 81: What's the use of having both blog and cms? Let's just stick with blog here

backend/templates/cms_index.html
```diff
@@ -0,0 +1,158 @@
...
+    </div>
+
+{% endblock %}
+
+
+
+
+
+
+
```
anastasia, line 99: Be mindful of the whitespace you're adding. Code should be orderly — it's harder to know what's going on with all the blank lines

backend/templates/cms_base.html
```diff
@@ -0,0 +1,75 @@
...
+    <meta name="theme-color" content="#000000"/>
+    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/css/bootstrap.min.css"
+          rel="stylesheet"
+          integrity="sha384-1BmE4kWBq78iYhFldvKuhfTAU6auU8tT94WrHftjDbrCEXSU1oBoqyl2QvZ6jIW3"
+          crossorigin="anonymous">
+    <link
+        rel="stylesheet"
+        href="https://cdn.jsdelivr.net/npm/bootstrap-icons@1.7.2/font/bootstrap-icons.css"
+    />
+    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.7.1/dist/leaflet.css"
```
anastasia, line 19: I don't believe you need most of these stylesheets adn scripts, definitely not leaflet (our mapping module)

backend/cms_app/models.py
```diff
@@ -0,0 +1,28 @@
...
+
+class Category(models.Model):
+    tag = models.CharField(max_length=20, help_text="Enter blog category")
+
+    def __str__(self):
+        return self.tag
+
+
+class Post(models.Model):
+    title = models.CharField(max_length=100)
```
anastasia, line 16: I don't think these tables are receiving any rows? I am seeing my posts being added to the cms built-in page table 
In the terminal: 
```
page = Page.objects.first()
page.get_page_title() 
```

anastasia:
Hey folks @Nisha-Nathan @JusticeV452! So I ended up removing django cms in favor of tinymce which is a rich text editor (a wysiwyg editor) on the backend and simple django templating on the front. Hopefully this will get us most of the way there without the pain of the giant django cms application. Let's discuss!

anastasia:
Here are the TODOs:
- [ ] What happens when there are no blog posts?
- [ ] What happens when a blog post is not published yet? Only signed in users should be able to see it
- [ ] Update admin panel with link/other stuff
- [ ] Add whatever helper methods we need on the model
- [ ] Style blog pages, remove all inline styles and add them to `blog.scss`
- [ ] Maybe move blog to React app?

PR #173, opened 2022-07-05
User Authentication

ADJohnson314:
This code will allow users to sign in to the website with a username and password, which will give them access to certain features such as editing blog posts. There is currently a sign in page, which has yet to reach the intended level of functionality.


cms_app/cms_app/asgi.py
```diff
@@ -0,0 +1,16 @@
+"""
```
anastasia, line 1: This shouldn't be necessary as we're using wsgi (find it in config/wsgi.py)

cms_app/cms_app/settings.py
```diff
@@ -0,0 +1,173 @@
+"""
```
anastasia, line 1: I don't think we want cms_app level-settings, we want to use the same settings we have in our config folder

PR #174, opened 2022-07-06
Map page frontend

qujohn314:
(WIP) This is the frontend implementation of the new map page based off of the final Figma designs. Also includes a new component for designing/implementing the project logo (blue and/or red squares). The page is contained in the MapPageView.js file and makes use of the new ParisMap and Logo components. There is currently duplicated code (for initializing and displaying the Paris Map) between the HomePage and MapPage as I was not sure whether the MapPage and HomePage are separate or one and the same. 

Completed Tasks:
-Map Page upon entry implemented
-Map Page adjusted for all screen sizes
-display links to other pages when map square or arrondissement clicked

To-Do Tasks:
-Add additional information based off clicked map square or arrondissement. Add real data / information.


frontend/pages/views/MapPageView.js
```diff
@@ -8,7 +8,17 @@ import LoadingPage from "../LoadingPage";
 import Logo from "../../components/Logo";
```
anastasia, line 8: There are a few unused imports on this page that should be cleaned up

frontend/components/Logo.js
```diff
@@ -9,16 +9,17 @@ export class Logo extends React.Component {
...
     }
 
     render() {
         return (
-            <div className="parent-bounding-box" style={{top: this.top+"%",
+            <div className={"parent-bounding-box-"+this.logo_type} style={{top: this.top+"%",
                                                          left: this.left+"%",
                                                          position:this.position_type}}>
-                <div className="bottom-right-red-rectangle"/>
-                <div className="top-left-blue-rectangle"/>
```
anastasia: Legend imported but never used. I'm also not seeing the logo anywhere? @qujohn314 can you add a screenshot of what you are seeing?

Thank you!

qujohn314: Capture3


Capture

Here's where it's used right now. I got rid of the side logo appearing in the bottom right corner since it was scrapped from the final design. The logo can change based on the screen size.

Capture2

anastasia:
Hey @qujohn314, I removed the extra NavBar in MapPageView. Is this PR ready to go?

qujohn314:
@anastasia Added some final changes. The dataset containing which map squares belong to which arrondissements was added as well as functionality displaying arrondissements info and links to to specific map squares. The PR is ready to go 👍

qujohn314:
May need to reopen/recreate PR in the future once more details on the final design is given and additional info/data is added, this PR gives a good idea of the mock design/functionality though.

frontend/scss/mapPageView.scss
```diff
@@ -0,0 +1,82 @@
+.page-style{
+    font-family: "eina_regular";
+}
+
+.page-body{
+    background-color: white;
+    max-height: 100%;
+    background-color: white;
```
anastasia, line 8: repeated line, delete

frontend/pages/views/MapPageView.js
```diff
@@ -0,0 +1,285 @@
...
+
+    render() {
+        if (!this.state.mapData || !this.state.arrondissementData || !this.state.filledMapSquares) {
+            return (<LoadingPage/>);
+        }
+
+        return (<div>
+                <Container fluid className="page-style">
+                    <Row className="page-body">
+                        <Col md={12} lg={7} className="m-0 p-0 min-vh-100" id="map-box">
```
anastasia, line 211: This map is currently eating the rest of the page in smaller screen sizes (`md={12}`). I know we want a full-width map but we also want to be able to navigate to the content below it. This might mean two things: first, showing that there IS content below by only allowing the map to take up 70% of the height of the screen instead of all of it. Also setting the `scrollWheelZoom` attribute to false on ParisMap would help (see editable example here: https://react-leaflet.js.org/).

backend/data/arrondissements_map_squares.json
```diff
@@ -0,0 +1,97 @@
+{"arrondissements":[{
```
anastasia: Awesome work. Thank you.

backend/app/views.py
```diff
@@ -402,6 +402,26 @@ def get_arrondissements_geojson(request, arr_number=None):
...
+    """
+    json_path = os.path.join(settings.BACKEND_DATA_DIR, 'arrondissements_map_squares.json')
+    with open(json_path, encoding='utf-8') as f:
+        data = json.load(f)
+
+    if arr_number is not None:
+        # Get data for a single unique arrondissement
+        data['arrondissements'] = [data['arrondissements'][arr_number-1]]
+
+    print(data['arrondissements'][0]["map_square_numbers"])
```
anastasia, line 421: remove print statements

frontend/components/Logo.js
```diff
@@ -0,0 +1,47 @@
+import React from "react";
```
anastasia, line 1: This isn't exactly the logo — it's just a decorator for our titles. Maybe calling it something like TitleDecorator would be more descriptive?

frontend/pages/views/MapPageView.js
```diff
@@ -0,0 +1,285 @@
...
+    // Creating 5 buckets based on lowest to highest number of photos per square
+    const twentyPctMax = Math.round(0.2 * maxNumOfPhotos);
+    const fortyPctMax = Math.round(0.4 * maxNumOfPhotos);
+    const sixtyPctMax = Math.round(0.6 * maxNumOfPhotos);
+    const eightyPctMax = Math.round(0.8 * maxNumOfPhotos);
+    const buckets = [0, twentyPctMax, twentyPctMax + 1,
+        fortyPctMax, fortyPctMax + 1,
+        sixtyPctMax, sixtyPctMax + 1,
+        eightyPctMax, eightyPctMax + 1, maxNumOfPhotos];
+    const layer = (<>
```
anastasia, line 62: no need to initialize variable if you're just going to return it. Can return this whole clause instead.

frontend/pages/views/MapPageView.js
```diff
@@ -0,0 +1,285 @@
...
+                const coords = mapSquareData.topLeftCoords;
+                const numberOfPhotos = mapSquareData.num_photos;
+
+                const mapSquareBounds = [
+                    [coords.lat, coords.lng],
+                    [coords.lat - MAPSQUARE_HEIGHT, coords.lng - MAPSQUARE_WIDTH]
+                ];
+                const link = "/map_square/" + index;
+                let mapSquareBucket = "";
+                // set of conditionals to calculate photo density for heat map
```
anastasia, line 77: Just FYI: If we were to keep this, this would have to be refactored out into its own component/set of functions. Otherwise we're just repeating ourselves. 
However, @ADJohnson314 is currently working on making this calculation on the backend so that we only have to calculate this once as opposed to every time the map loads.

frontend/components/ErrorNotFoundComponent.js
```diff
@@ -1,18 +0,0 @@
-import React from "react";
```
anastasia, line 1: Is there a reason this file was deleted? We should probably keep it.

anastasia (changes requested):
Hey — great work @qujohn314. I left comments, mostly pedantic (but would still love for you to fix), some more urgent. Once they are done I will merge!

frontend/components/ErrorNotFoundComponent.js
qujohn314: I am pretty sure it got added automatically when I accidentally deleted a component I was working on. It caused a bunch of errors and reformatted some code until I added the original file back and deleted this one.

qujohn314:
I addressed all of the comments. One thing I was stuck on was making the map scroll zoom functionality disabled on smaller viewports while keeping on larger view port sizes. I got it to work, however now there is a delay in the map being displayed on the page. The text appears first and then the map rather than them both appearing simultaneously. After doing a lot of reading, I believe it could be that I need to lift some state handling/changing from the paris map to the map page view, but I was unsure how to do this exactly.

Once I address this lag/delay problem either with your assistance or I just disable the scroll zoom functionality for all viewport sizes, this PR should be good to go.

frontend/pages/views/MapPageView.js
```diff
@@ -198,30 +204,40 @@ class MapPage extends React.Component {
         } catch (e) {
             console.log(e);
         }
+
+        this.updateViewport();
+        window.addEventListener("resize", this.updateViewport);
```
anastasia, line 209: This is great logic. The only thing that's missing that will speed things up is a debouncer (a function that slows down how often the inner function, `this.updateViewport` is called). Let's add that to a future/following TODO

frontend/scss/variables.scss
```diff
@@ -2,3 +2,10 @@ $color-turquoise: #20CCD7;
 $color-orange: #FB2F2A;
 $color-mid-gray: #5E5F5F;
 $color-link: #3295B8;
+
+.link{
+    font-weight: 500;
+    text-decoration: underline;
+    color: $color-link;
+    display: inline;
+}
```
anastasia, line 11: 🎉

anastasia:
Great, thank you for addressing all the comments. See the note about the debouncer. Want to take that on after?

qujohn314:
Yeah I will add that to my TO-DO list 👍

PR #176, opened 2022-07-11
updated Django admin to make Photo and MapSquare list view more readable

czheng10:
Adds PhotoAdmin and MapSquareAdmin to admin.py 

The list view for Photos now displays:
- ID
- Number
- Thumbnail
- Map Square Number/ID
- Photographer ID and Name
- Librarian Caption
- Photographer Caption

Slides are added to edit view

The list view for MapSquares now displays:
- ID
- Number
- Coordinates
- Photo Count
- Boundaries

All photos of that map square are added to edit view

Searchbar added for both


backend/app/admin.py
```diff
@@ -13,10 +17,118 @@
...
+            return
+    photographer_info.short_description = 'Photographer'
+
+    def photo_thumbnail(self, obj):
+        """
+        Displays thumbnail of photo obj
+        """
+
+        # taking photos from local dir
+        obj_path = os.path.join(settings.LOCAL_PHOTOS_DIR, str(obj.map_square.number),
```
anastasia, line 109: I think we want this to go to the THUMBNAIL_DIR (which we don't have yet). We definitely don't want to pull in all full-sized images into our admin panel

backend/app/admin.py
```diff
@@ -13,10 +17,118 @@
...
+        # filter for all photo objects of obj map square
+        photo_obj = Photo.objects.filter(map_square__number=obj.number)
+        for photo in photo_obj:
+            # path to photo
+            obj_path = os.path.join(settings.LOCAL_PHOTOS_DIR, str(obj.number),
+                                    str(photo.number)+'_photo.jpg')
+
+            # path to photo's admin change page
+            photo_path = os.path.join('/admin/app/photo', str(photo.id))
+            cmd += '<a target = "blank" href = "{photo_url}"> <img src="{url}" ' \
```
anastasia, line 62: This is very (!!) minor but I wouldn't use the variable `cmd` here because this isn't a command — it's a link! Maybe html = '<a target="blank"....' etc

anastasia: Also no spaces around attribute setting in html is nicer :-) Like this:
<a target="blank" href="{photo_url}">

backend/app/admin.py
```diff
@@ -1,8 +1,12 @@
 """
 This file controls the administrative interface for paris_1970 app
 """
-
+from django.conf import settings
+from PIL import Image
```
anastasia, line 5: We're not using this package or `reverse`, so can remove

anastasia (changes requested):
Hey — this looks great in general! Added some very small nitpicks. Ready to be a real PR

anastasia:
Thanks @czheng10! Good to go for now (with possible thumbnail changes in the future).

PR #177, opened 2022-07-11
Script to Rename Photos to x_slide and x_photo

IsmailA23:
This script renames photos from BHVP_PH_CetaitParis_DP_1538_01_015 and BHVP_PH_CetaitParis_DP_1538_01_016.jpg to 16_slide.jpg and 16_photo.jpg

Images come in two versions: as a slide (image with the surroundings/uncropped), and as a photo (the image cropped and with no surroundings). 

Example:

1_slide copy

1_photo copy

These two photos used to be named BHVP_PH_CetaitParis_DP_1538_01_001.jpg and BHVP_PH_CetaitParis_DP_1538_01_002.jpg. This script changes their names to 1_slide.jpg and 1_photo.jpg

It also creates a csv file that keeps track of the changes.


.github/scripts/Rename_Photos
```diff
@@ -0,0 +1,26 @@
+#This script renames photos from BHVP_PH_CetaitParis_DP_1538_01_015 and BHVP_PH_CetaitParis_DP_1538_01_016.jpg to 16_slide.jpg and 16_photo.jpg
+
+import math
+import os
+import csv
+
+folder = r"/Users/ismailassafi/Documents/GitHub/paris_1970/assets/images/photos/1538/"
```
anastasia, line 7: This path shouldn't be hardcoded! We will probably have a different a different path to input

anastasia: Take a look at @czheng10's scripts in this folder and how they deal with this: https://github.com/dhmit/paris_1970/blob/main/backend/scripts/imageconversion.py

.github/scripts/Rename_Photos
```diff
@@ -0,0 +1,26 @@
...
+count = 1
+
+for file_name in os.listdir(folder):
```
anastasia, line 10: Take a look at the `tqdm` package, also used by @czheng10 here: https://github.com/dhmit/paris_1970/blob/main/backend/scripts/imageconversion.py#L36
This will allow the user running this script to see progress while iterating

.github/scripts/Rename_Photos
```diff
@@ -0,0 +1,26 @@
...
+    else:
+        type_of_img = "slide"
+
+    source = folder + "/" + file_name
+    destination = folder + "/" + str(p) + "_" + type_of_img + ".jpg"
+    os.rename(source, destination)
+    count += 1
+    with open('dictionary.csv', 'a+') as csvfile:
+        csvwriter = csv.writer(csvfile)
+        csvwriter.writerow([source, destination])
```
anastasia, line 26: Let's add a print message for the user which will include the destination folder as well. This will let the user know that the job is over and where to find the new photos.

.github/scripts/Rename_Photos
anastasia: The function should take in a source dir and a destination dir as arguments.

anastasia (changes requested):
Hi @IsmailA23, including some requests for changes here. Congrats on your first PR! 🚀

.github/scripts/Rename_Photos
```diff
@@ -0,0 +1,26 @@
+#This script renames photos from BHVP_PH_CetaitParis_DP_1538_01_015 and BHVP_PH_CetaitParis_DP_1538_01_016.jpg to 16_slide.jpg and 16_photo.jpg
```
anastasia, line 1: Hey — remember that this file needs to live in the scripts directory inside of our project — not .github folder

.github/scripts/Rename_Photos
anastasia, line 7: I think this is the old file?  I remember seeing another file with a function with arguments to call instead of a hard-coded folder path

IsmailA23: Yes, sorry. Let me change this right now.

PR #179, opened 2022-07-12
User groups and login/logout pipeline

Nisha-Nathan:
I made a signup feature that allows users to register and be assigned specific permissions that allow them to add, edit or remove blog posts in the admin panel. I also made a login feature that can be used by both the user and admin to access the admin page with different levels of permissions. Lastly, I redirected the log-out URL in the admin page to the home page of the site instead of the default page that prompts login again.

@anastasia edit: 
Added blog-only permissions set. In Admin, change user's permissions by adding `blog_permissions` group

`python manage.py migrate`


anastasia:
I believe there are some requirements missing from requirements.txt (django-crispy-forms and crispy-bootstrap5). Thanks for the note about installing!

backend/blog/migrations/0003_auto_20220713_1105.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 3.2.14 on 2022-07-13 15:05
+
+from django.db import migrations, models
+import taggit.managers
+
+
+class Migration(migrations.Migration):
+
```
anastasia, line 8: Great, thanks for adding this!

backend/blog/views.py
```diff
@@ -25,3 +33,45 @@ def blog_post(request, slug):
     tags = Tag.objects.all()
     context = {'post': post, 'tags': tags}
     return render(request, 'blog/post.html', context)
+
+
+def user_register(request):
+    if request.method == "POST":
```
anastasia, line 39: Hey – we should talk about this. This actually opens us up to anyone being able to register and therefore add/delete/edit blog posts. This is very dangerous! I appreciate all the effort that you put in this PR. 
Like we talked with @ADJohnson314 I think Django does a really great job with the bare bones registration that exists already. I was expecting only some permissions settings, which you're doing on any new user registration, but you're also doing so much more! Great exploration but I think this means we need to rethink our set up.

anastasia (changes requested):

backend/config/urls.py
```diff
@@ -80,5 +81,8 @@ def react_view_path(route, component_name):
     # blog urls
     path('blog/', blog_views.index, name="blog_home"),
     path('blog/<str:slug>/', blog_views.blog_post,
-         name='blog-detail')
+         name='blog-detail'),
+    # sign up urls
+    path('login/',auth_views.LoginView.as_view(template_name='admin/login.html'),name='login'),
```
anastasia, line 86: great!

backend/app/migrations/0002_auto_20220718_1614.py
```diff
@@ -0,0 +1,37 @@
+# Generated by Django 3.2.14 on 2022-07-18 16:14
+
+from django.db import migrations
+
+
+def permissions_func(apps, _):
```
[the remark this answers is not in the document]
Nisha-Nathan: great!

PR #180, opened 2022-07-12
Blog admin tweaks

JusticeV452:
- Added tag field to BlogPost model
- Added help text to slug field
- Added `taggit` to Django `INSTALLED_APPS`

This PR requires python manage.py migrate (taggit added to apps)

image

### DIFF
--- a/backend/blog/admin.py
+++ b/backend/blog/admin.py
@@ -1,8 +1,5 @@
 from django.contrib import admin
-# from django import forms
-from django.contrib.flatpages.admin import FlatPageAdmin
 from .models import BlogPost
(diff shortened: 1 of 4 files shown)


anastasia:
Hey @JusticeV452 looks great! Thank you for fixing this. I added one more fix to my code but worth noting — 
we're now only generating the slug if it isn't defined (therefore I changed the help text on the `slug` field to "Auto generated from title field if not defined.")
This will allow blog authors to manually set the slug if it's not to their liking, also. We should probably in the future make sure that the slug field is unique, as slugs form our blog URLs. Let's punt on that though :-)

PR #181, opened 2022-07-15
Hide Unpublished Blog Posts

JusticeV452:
- Users will see 404 page with navbar and link to the blog home page if they attempt to access a blog post that:
  - Does not exist
  - Is unpublished and they are not the author of the post
- Tags associated only with unpublished blog posts are not shown on the blog home page
- Implements the `get_absolute_url` method for BlogPost model

Blog home:
image

Blog post that does not exist:
image

Unpublished blog post, logged in:
image

Unpublished blog post, not logged in:
image


anastasia:
Great work, thank you!

A near-future `TODO` (@JusticeV452 if you feel like taking this on or I can assign it to someone else): users that have blog editing privileges should be able to see their blogposts in preview mode. That is, if they navigate to their blog's URL, they should see their blog along with some kind of notification like this at the top:
```THIS IS A PREVIEW ONLY. TO SEE THIS LIVE, CLICK "published" IN THE ADMIN PANEL```

File: backend/blog/views.py [hunk 1]
```diff
@@ -21,7 +20,9 @@ def blog_post(request, slug):
     """
     Single blog post view
     """
-    post = BlogPost.objects.get(slug=slug)
-    tags = Tag.objects.all()
-    context = {'post': post, 'tags': tags}
+    posts = BlogPost.objects.filter(slug=slug)
+    if not posts or (not posts[0].published and request.user != posts[0].author):
```
anastasia, line 24: This is a great direction, but let's make this even simpler. If user is signed in with blog privileges, they can see the post. The reason I think this rather than just author: students might want to share their unpublished posts with Catherine or someone else. Certainly sharing a URL will be easier than telling them where to go to the admin panel to see their work in progress. How does that sound?

backend/blog/views.py
```diff
@@ -21,7 +20,9 @@ def blog_post(request, slug):
...
+        return render(request, 'blog/blog_404.html')  # Respond with 404 page
+    tags = posts[0].tags.names()
+    context = {'post': posts[0], 'tags': tags}
     return render(request, 'blog/post.html', context)
```
anastasia, line 28: In post.html, let's add a banner at the top if the post is unpublished.
Something like:
```THIS IS A PREVIEW ONLY. TO SEE THIS LIVE, CLICK "published" IN THE ADMIN PANEL```

**Review by anastasia — CHANGES_REQUESTED** [review]
Almost ready to go! A couple structural and cosmetic changes

File: backend/blog/views.py (hunk 1 above)
Comment (anastasia) [reply backend/blog/views.py:24]: something like that! If they don't have permissions then send them to 404, otherwise send them the blog page (with the banner)

File: backend/blog/views.py [hunk 3]
```diff
@@ -21,8 +21,14 @@ def blog_post(request, slug):
...
-        return render(request, 'blog/blog_404.html')  # Respond with 404 page
-    tags = posts[0].tags.names()
-    context = {'post': posts[0], 'tags': tags}
-    return render(request, 'blog/post.html', context)
+    if posts and (
+        posts[0].published
+        or request.user == posts[0].author
+        or request.user.is_staff
+        or request.user.is_superuser
+    ):
```
anastasia, line 29: nice

backend/templates/blog/post.html
```diff
@@ -1,6 +1,12 @@
 {% include "base.html" %}
 {% block content %}
     <div class="blog" id="app_root">
+        {% if not post.published %}
+            <h5 style="white-space: pre-wrap">
```
anastasia, line 5: nice!

anastasia:
Nice, nice. Ready to merge!

PR #182, opened 2022-07-19
Updating database with photographers

nfrieden:
Includes csv with new photographer data and python script used to upload said data into SQL


backend/scripts/sqliteversion.py
```diff
@@ -0,0 +1,22 @@
+import sqlite3
```
anastasia, line 1: Nice. Can you rename this file to something like "updatephotographers_sqlite.py"? That way it'll be more clear what it is

anastasia (changes requested):
Great! Just one small change requested

backend/scripts/sqliteversion.py
nfrieden: Done!

anastasia (approved):
Great, thanks!

anastasia:
🎉

PR #183, opened 2022-07-20
Backend tests

awu482519:
Developing backend tests for api endpoints in backend/app/tests.py, divided into analysis-related functionality and analysis-unrelated functionality. Includes minor changes to models.py and serializers.py for more flexibility i.e. returning empty values when map_square_coords does not exist and allowing images to be potentially taken from different local source directories.


backend/config/settings/base.py
```diff
@@ -22,7 +22,8 @@
 ANALYSIS_DIR = Path(PROJECT_ROOT, 'backend', 'app', 'analysis')
 ANALYSIS_PICKLE_PATH = Path(BACKEND_DIR, ANALYSIS_DIR, 'analysis_results')
 LOCAL_PHOTOS_DIR = "/static/images/photos"
-LOCAL_SRCS_DIR = Path(PROJECT_ROOT, 'assets', 'images', 'photos')
+# LOCAL_SRCS_DIR = Path(PROJECT_ROOT, 'assets', 'images', 'photos')
+LOCAL_SRCS_DIR = Path(PROJECT_ROOT, 'backend', 'data', 'test_photos')
```
anastasia, line 26: You have changed `LOCAL_SRCS_DIR` to point to test photos directory and added this as the default value to non-test functions. Is this correct?

anastasia (changes requested):
@awu482519 Great work! Some small changes/explanations requested

backend/app/models.py
```diff
@@ -81,6 +80,29 @@ def get_image_data(self, as_gray=False, use_pillow=False):
...
         unique_together = ['number', 'map_square']
 
+    def get_image_local_filepath(self, src_dir=settings.LOCAL_SRCS_DIR):
+        """
+        Get the image local filepath if it exists.
+
+        We try for a local filepath first, as that's faster,
+        and we fallback on Google Drive if there's nothing local.
+        """
+        if self.cleaned_src:
```
anastasia, line 90: Lines 90 through 96 don't serve any purpose. Can be deleted!

backend/config/settings/base.py
anastasia: Looking at `TEST_PHOTOS_DIR` value, I think this is a mistake.

awu482519: yes!

awu482519: as in, I did intentionally change LOCAL_SRCS_DIR to have the same value as TEST_PHOTOS_DIR. should I just say LOCAL_SRCS_DIR = TEST_PHOTOS_DIR instead? (after having defined TEST_PHOTOS_DIR of course)

backend/app/tests.py
```diff
@@ -1,22 +1,194 @@
...
+        # create 3 map squares
+        for i in range(3):
+            map_square = MapSquare.objects.create(number=i + 1, coordinates='24, 25')
+            photographer = Photographer.objects.create(map_square=map_square, number=i + 1,
+                                                       name=names[i])
+            path = os.path.join(settings.TEST_PHOTOS_DIR, f'{i + 1}')
+            os.mkdir(path, mode=0o755)
+            # in each map square, create 4 empty photos
+            for j in range(4):
+                photo = Photo.objects.create(number=j + 1, map_square=map_square,
```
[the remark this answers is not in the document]
awu482519: i use the photo variable here as input for the PhotoAnalysisResult objects. Otherwise, I made all the changes requested!

backend/config/settings/base.py
anastasia: hmm this doesn't seem right. What is the original use of LOCAL_SRCS_DIR? We don't want regular functions (like analysis functions) to default to test directories, we want them to look at the real local photos

awu482519: makes sense, I've changed it back so LOCAL_PHOTOS_DIR (the same as LOCAL_SRCS_DIR) is differentiated from TEST_PHOTOS_DIR and the latter is passed into the test cases as needed

anastasia:
Great work 🚀 !

awu482519:
thank you for all the help :relaxed:

PR #185, opened 2022-07-20
Adj mapsquares

ADJohnson314:
The current goal is to have the script import the MapSquare model and run MapSquare.objects.all(), so that it can gather the data for every map square and dump it into a JSON file. The Home.js file (or any other file) would then retrieve this file instead of making a call to the all_map_squares view. The issue right now is that the import from models.py isn't working.


anastasia:
Hey — can you pull from `origin` `main` and try again? I added an __init__ file to scripts which should make importing possible

ADJohnson314:
So the init file worked, and I was able to import what I needed from models.py. Now there seems to be a problem with the settings file. The error says: "django.core.exceptions.ImproperlyConfigured: settings.DATABASES is improperly configured. Please supply the ENGINE value. Check settings documentation for more details." But the ENGINE value is already supplied in DATABASES in base.py. I made sure the NAME value is there as well, but that doesn't seem to help.

backend/app/models.py
```diff
@@ -15,6 +15,21 @@
...
+    a list of foreign keys to all the Photos that are in this Map Square, and
+    the boundary of this Map Square.
+    """
+    name = models.CharField(max_length=252)
+    number = models.IntegerField(null=True)
+    boundaries = models.CharField(max_length=252)
+    coordinates = models.CharField(max_length=252)
+
+    class Meta:
+        app_label = 'user_app'
```
anastasia, line 30: I am not sure what's going on here! Is there a reason you are adding `app_label = 'user_app'`? It should be in app

backend/app/models.py
```diff
@@ -80,18 +95,7 @@ def get_image_data(self, as_gray=False, use_pillow=False):
...
-    """
-    This model contains data about a specific Map Square, which includes its name,
-    a list of foreign keys to all the Photos that are in this Map Square, and
-    the boundary of this Map Square.
-    """
-    name = models.CharField(max_length=252)
-    number = models.IntegerField(null=True)
-    boundaries = models.CharField(max_length=252)
-    coordinates = models.CharField(max_length=252)
+        app_label = 'user_app'
```
anastasia, line 98: This is not needed

backend/app/models.py
ADJohnson314: Before I added that line, there was an error that said "Model class app.models.MapSquare doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS." This line seemed to resolve the error.

PR #187, opened 2022-07-22
creation and checking uniqueness of slugs

Nisha-Nathan:
I worked on checking the uniqueness of slugs where an error is thrown when a user writes a slug that is not unique and  a user is given an option to leave the slug field blank as a slug will be autogenerated from the blog title and current date when a blog is saved or published


backend/blog/models.py
```diff
@@ -8,14 +8,17 @@
...
+
 
 class BlogPost(models.Model):
     author = models.ForeignKey(User, null=True, on_delete=models.DO_NOTHING)
     title = models.CharField(max_length=200, null=False, blank=False)
     subtitle = models.CharField(max_length=1000)
     slug = models.SlugField(
-        max_length=100, blank=True, null=True, help_text=_("Auto generated from title field "
-                                                           "if not defined.")
+        max_length=100, blank=True, null=True, unique=True, help_text=_("Auto generated from title "
```
anastasia, line 19: hmmm.... this unique constraint does not seem to be working

backend/blog/models.py
```diff
@@ -32,11 +32,25 @@ def get_absolute_url(self):
...
+        slug = self.cleaned_data['slug']
+
+        if BlogPost.objects.filter(title=title).exists():
             date = datetime.datetime.today()
-            self.slug = '%s-%i-%i-%i' % (slugify(self.title), date.year, date.month,
-                                         date.day)
-        elif not self.slug:
-            self.slug = slugify(self.title)
-        super(BlogPost, self).save(*args, **kwargs)
+            self.cleaned_data['slug'] = '%s-%i-%i-%i' % (slug, date.year, date.month,
```
anastasia, line 49: Hey — this is almost there but the logic isn't sound.
I have two blog posts, both called "The Blogpost". 
Post A gets created fine, the slug is "the-blogpost"
Post B errors out on creation (great!) and I manually set "the-blogpost-1" and hit save.
Instead, the slug is overwritten as "the-blogpost-1-[today's date]"
We should not be overwriting the user's unique set slug!

anastasia:
Woohoo! 🚀

PR #188, opened 2022-07-25
Cleaning Up Repo, Creating New Paths, TAG PAGE

czheng10:
Organizes the paths in urls.py, deleting paths that will no longer be used in the new website, as well as adding new paths for new pages that now exist. Similarly, new methods added to views.py, as well as removing methods may no longer be needed. base.html navbar and index.js also updated to match changes.

To get tags to work, run
`python manage.py runanalysis yolo_model`
in your terminal.

You will now be able to see live tags on the PhotoView page (they may or may not be accurate to the picture, but that is a database issue!). If you click on a tag, it will take you to a TagView page that displays the correct photos from the analysis.

Note: The JS files for unused pages (ex. AnalysisView.js) still exist in the repo


anastasia:
@czheng10 Is this ready to go?

czheng10:
> @czheng10 Is this ready to go?

@anastasia yep, just communicating with @nfrieden on making this a pull request.

frontend/pages/Tags.js
```diff
@@ -0,0 +1,78 @@
...
+        super(props);
+        this.state = {
+            loading: true,
+            tagData: null
+        };
+    }
+
+    async componentDidMount() {
+        try {
+            const response = await fetch(`/api/tag/${this.props.tagName}/`);
```
anastasia, line 18: This shouldn't be a separate API call. We can get this information on the backend when the view is called, since we're passing in the tag we're searching for anyway. The view should return necessary information (photos that include tag item)

frontend/pages/Tags.js
```diff
@@ -0,0 +1,78 @@
+import React from "react";
```
anastasia, line 1: Let's relabel this file to Tag or TagView, since we are not using it to render a list of tags, but to show just one tag

anastasia:
Please add the necessary commands to run to make your code work in your description of the PR. Thanks @nfrieden for the heads up!

frontend/pages/Tags.js
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+
+    formatTag(tagName) {
+        let array = tagName.split(" ");
+        for (let i = 0; i < array.length; i++) {
+            array[i] = array[i].charAt(0).toUpperCase() + array[i].slice(1);
+        }
+        let result = array.join(" ");
+        return result;
```
anastasia, line 39: no need to add variable and then return, can just return array.join(" ")

frontend/pages/Tags.js
```diff
@@ -0,0 +1,78 @@
...
+            }
+        } catch (e) {
+            console.log(e);
+        }
+    }
+
+    formatTag(tagName) {
+        let array = tagName.split(" ");
+        for (let i = 0; i < array.length; i++) {
+            array[i] = array[i].charAt(0).toUpperCase() + array[i].slice(1);
```
anastasia, line 36: I believe you can just accomplish this with a CSS property:
`text-transform: capitalize;` Let's remove this logic

frontend/pages/views/PhotoView.js
```diff
@@ -461,9 +461,11 @@ export class PhotoView extends PhotoViewer {
                         </div>
 
                         {tagList.map((word) => (
-                            <button className="tag-button" key={word.id}>
-                                {word}
-                            </button>
+                            <a key={"tag"} href={`/tag/${word.toLowerCase()}/`}>
```
anastasia, line 464: no need to lowercase if we're later title casing these, let's just use CSS to do this instead

frontend/pages/views/PhotoView.js
```diff
@@ -10,7 +10,7 @@ import {Dropdown, OverlayTrigger, Popover, Modal} from "react-bootstrap";
 import ExpandIcon from "../../images/expand.svg";
 import QuestionIcon from "../../images/question.svg";
 
-let tagList = ["Construction", "People", "Building"];
+let tagList = ["Bicycle", "Stop Sign", "Cow"];
```
anastasia, line 14: Want to add the live tags here instead?

anastasia (changes requested):
Hey — good work @czheng10 and @nfrieden! A few changes requested. Let's add the live tag logic to the photo page if possible. Come find me if my comment about removing the separate API call doesn't make sense (or anything else, of course)!

anastasia:
Excellent work @czheng10 and @nfrieden! Merging

PR #189, opened 2022-07-25
Debouncer added to resize event

qujohn314:
Debouncer added to resize event to speed up component loading times


anastasia (changes requested):
I guess this works But! 
1. You added lodash but did not update `package.json` with this new package
2. You could achieve the same result without adding a whole package by just adding some common JS. See an example here: https://codepen.io/ondrabus/pen/WNGaVZN (we have common.js for exactly this)

qujohn314:
Debounce method moved to common.js 👍

PR #191, opened 2022-07-29
Add event listener for escape key

JusticeV452:
- Enables the user to close the fullscreen photo popup on the `PhotoView` page by pressing the Escape key


anastasia (approved):
Woohoo!

PR #196, opened 2022-08-03
Search page scaffold

JusticeV452:
- Changed search API view to a GET view
- Changed search API view to return search results based on a string of keywords separated by spaces 
    - Search keywords with spaces should be input surrounded by quotes (Ex: person car "Paris 13E")
    - Returns only photos that match all keywords in some way (photographer_number=keyword or caption contains keyword)
    - Results are sorted in order of descending max yolo_model confidence for searched keyword (if keyword is not a valid yolo_model tag, confidence defaults to 100)
- Implement basic structural style elements from the Figma into the frontend
    - Enter tags separated by space and press Enter key or "Advanced Search" to show search results

Screenshots:
Default
image

person
image

person and car
image

string "person car"
image

photographer number (17602)
image


anastasia:
Hey @JusticeV452 *excellent* work!!

PR #198, opened 2022-08-04
new UROP display for About page

nfrieden:
About page UROP view is now much closer to how it appears in the wireframe. Hovering over a UROP/staff member's picture will display their name and a paragraph about them. 

To-do: 
- Add alternating blue/red colors for background on hover
- Have UROPs write a paragraph about themselves to display


anastasia:
Thank you!

PR #207, opened 2022-09-06
Development settings now use remote database and image bucket

ryaanahmed:
In advance of fall 2022 UROP - we're moving to using a single remote database and an S3 bucket containing the photos for development. 

[edit: 2022-09-08]: Ended up putting more changes to the dev experience in here than I'd expected. In addition to moving to the remote database and image hosting, this also includes installation scripts for Windows and macOS/Linux to make the installation process easier.


ryaanahmed:
TODO before merging this in: add CI db settings that spin up a sqlite db instead.

PR #208, opened 2022-09-16
Image conversion script now operates over the whole dataset

ryaanahmed:
Previously, the image conversion script ran on a single directory of images. The actual use case is that we have a gigantic containing folder of images, so this PR changes that script so it runs over that directory and batch processes everything inside of it.

It also switches us from `subprocess.run` to `subprocess.Popen` in order to parallelize the processing.

NOTE: I pretty much hard coded the batch size for subprocess.Popen because I assume I'm the only person running this script in the lab, and I tweaked all of the values to be good for my home desktop machine (12-core Ryzen). So, it might not work on yours.


backend/scripts/imageconversion.py
```diff
@@ -1,65 +1,65 @@
...
-            [ex. 'Users/bob/Desktop/new/']
-    :param old_filetype: (str) image format of input images
-            [ex. 'TIF']
-    :param new_filetype: (str) desired output image format
-            [ex. 'jpg']
-    :param commands: (list) commands to apply to images, default = None
-            [ex. ['-quality, '20%']
+    :param src_dir_path: Path object to directory containing subdirectories of images
+    :param dest_dir_path: Path path object where output is intended to be saved in
+    :param commands: (list of strings) commands to apply to images
```
ryaanahmed, line 15: By the way @czheng10 -- I got rid of the two params specifying filetype bc we're always dealing with jp2 images.

PR #230, opened 2022-09-30
Changed nav background color from #1C1A1A to #000000

sharonyu524:
Changed the background color of the navigation bar to match the figma style guide.


ryaanahmed (changes requested):
Thanks! One thing to check before we merge this in: the SCSS variable `color-black` is used in a few places, not just the navbar. It's also used as the color for buttons (in `frontend/scss/styles.scss`) and as the border color for the legend (in `frontend/scss/legend.scss`).

Could you double check that changing this global variable keeps those styles as they should be? If not, we'll need to make a new variable.

Also: for style PRs, please include a screenshot of the changes, so we can see what's changed directly.

Thanks!

ryaanahmed:
@sharonyu524 and I talked about this -- the existing color was actually just barely-not-black, so changing it to black for sure doesn't mess with anything. Merging!

ryaanahmed (approved):

PR #233, opened 2022-09-30
Limit Similar Photos

irinazoccolini:
This PR limits the number of similar photos shown on the single photo page to 10. Instead of doing this on the frontend, it's now done on the backend.

### DIFF
--- a/backend/app/serializers.py
+++ b/backend/app/serializers.py
@@ -53,7 +53,11 @@ class PhotoSerializer(serializers.ModelSerializer):
         for analysis_result in analysis_results:
             name = analysis_result['name']
             result = json.loads(analysis_result['result'])
-            analyses_dict[name] = result
+            if name == "photo_similarity.resnet18_cosine_similarity":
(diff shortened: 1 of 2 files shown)


backend/app/serializers.py
```diff
@@ -53,7 +53,11 @@ def get_analyses(instance):
         for analysis_result in analysis_results:
             name = analysis_result['name']
             result = json.loads(analysis_result['result'])
-            analyses_dict[name] = result
+            if name == "photo_similarity.resnet18_cosine_similarity":
+                analyses_dict[name] = result[:min(10, len(result))]
```
ryaanahmed, line 57: Handy thing: slice syntax in Python actually reduces the stop value to the length of the collection if you ask for more values than the length.

So: no need for the call to `min` in this expression -- Python does it for you.

ryaanahmed (changes requested):
Looks good! See my comment for a small change.

ryaanahmed (approved):
thanks for the changes -- merging!

PR #234, opened 2022-10-07
Layout Changes to Photographers List Image Display

bukunmishodipo:
I created a folder in assets for the photographer's images, fixing the image links. Also edited the size of the images and the spacing to match what's specified in the Figma document.


backend/app/views.py
```diff
@@ -517,7 +517,8 @@ def photographer_list_view(request):
     """
     Photographer list page
     """
-    photos_dir = os.path.join(settings.LOCAL_PHOTOS_DIR, 'photographers')
+    # photos_dir = os.path.join(settings.LOCAL_PHOTOS_DIR, 'photographers')
```
ryaanahmed, line 520: Don't leave dead code in PRs -- if you've made a change, please do go ahead and delete the code that's replaced.

backend/config/settings/base.py
```diff
@@ -23,6 +23,7 @@
 ANALYSIS_PICKLE_PATH = Path(BACKEND_DIR, ANALYSIS_DIR, 'analysis_results')
 LOCAL_PHOTOS_DIR = "/static/images/photos"
 LOCAL_PHOTOS_LOCATION = Path(PROJECT_ROOT, 'assets', 'images', 'photos')
+PHOTOGRAPHERS_DIR = "/static/images/photographers"
```
ryaanahmed, line 26: When setting paths, generally don't set them as strings. Since paths are represented differently on Unix-based platforms and Windows, this code will crash on Windows. Check out the line above (setting `LOCAL_PHOTOS_LOCATION`) for how to use Python's pathlib to create cross-platform compliant paths.

It looks like earlier in the project, someone committed a similar problem with `LOCAL_PHOTOS_DIR` -- could you please fix this in this PR as well? I'm not sure that this path is used in the project anymore, though. If you find that it isn't, go ahead and delete it.

frontend/scss/photographerListView.scss
```diff
@@ -1,9 +1,30 @@
...
+    width: 13.7%;
 
 }
 
+.photographers-container{
+    padding: 1px;
+}
+
 .child{
+    margin-left: 10px;
```
ryaanahmed, line 13: Looks good. Just FYI -- one way you can do this in one line is:
`margin: VERTICAL HORIZONTAL`

so in this case
`margin: 20px 10px`

see https://developer.mozilla.org/en-US/docs/Web/CSS/margin for docs on this.

ryaanahmed (changes requested):
Nice work. I'm requesting changes for the Path thing (see my review comments). In general, when making a PR for a style change, please include a screenshot (ideally before and after) so we can easily see what's changed in GitHub.

PR #235, opened 2022-10-07
Adding search by name to photographers page

Menamonmon:
Created a new Django API endpoint to search the photographers by their name. Added a search bar on the front end that utilizes the endpoint and stores the photographers that match the search query as a state instead of capturing it from the props passed from the Django views. Currently, if no search query is given, the page only displays the first 50 photographers, otherwise, it displays all of the photographers who match the given query. 

This approach still has some inefficiencies because if you look up "a" the page will load all of the photographers with the letter "a" in them, but the next step after merging this pr would be to add pagination to these results so that the user can load them incrementally as they need to by scrolling to the bottom of the page. 

Below is how the search bar currently looks like. Planning on reimplementing it after the design is finalized and we implement search by photographer location on the backend. 

Screen Shot 2022-10-07 at 4 27 40 PM

### DIFF
--- a/frontend/pages/views/PhotographerListView.js
+++ b/frontend/pages/views/PhotographerListView.js
@@ -1,23 +1,39 @@
 import React from "react";
 import Footer from "../../components/Footer";
 import * as PropTypes from "prop-types";
-
+import { debounce } from "../../common";
 
 export class PhotographerListView extends React.Component {
     constructor(props) {
         super(props);
         this.state = {
-            photographers: JSON.parse(this.props.photographers)
+            photographers: [],
         };
     }
 
     hrefFunc(number) {
         return `/photographer/${number}`;
-    };
+    }
 
     srcFunc(number) {
         return `${this.props.photoListDir}/${number}_photo.jpg`;
-    };
+    }
+
+    updatePhotographers(name) {
+        debounce(async () => {
+            const fetchedPhotographers = await this.fetchPhotographers(name);
+            this.setState({ photographers: fetchedPhotographers });
+        }, 300)();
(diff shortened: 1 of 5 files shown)


frontend/pages/views/PhotographerListView.js
```diff
@@ -1,68 +1,97 @@
...
         return (
             <>
                 <div className="row">
-                    <p style={{
-                        fontSize: "30px",
-                        fontWeight: "600"
-                    }}>Photographers</p>
+                    <p
+                        style={{
+                            fontSize: "30px",
```
ryaanahmed, line 68: Although this isn't the point of this PR, since you're working in this area -- this style info should get abstracted out into an SCSS file at some point. (doesn't have to happen in this PR, but before we finalize this module)

frontend/pages/views/PhotographerListView.js
```diff
@@ -1,68 +1,97 @@
...
+                            fontWeight: "600",
+                        }}
+                    >
+                        Photographers
+                    </p>
+                </div>
+                <div>
+                    Search By Name:{" "}
```
ryaanahmed, line 76: I think maybe you were trying to achieve a non-breaking space here `&nbsp;`

ryaanahmed (changes requested):
Great work! See my comments within for some changes -- mostly code style stuff. And thanks for fixing the linter problems!

Menamonmon:
@ryaanahmed I resolved most of the comments except for the styling ones which I will complete along with the rest of the styling and design for the entire search function in a future pr.

PR #237, opened 2022-10-14
Added columns to Photographer List View, added dunder str methods to models

czheng10:
Added the following columns to Photographer Admin List View:
- ID
- Name
- Number
- Approx Location
- Map Square

Added __str__ methods to Photo, Photographer, Map Square, Analysis Result Models


frontend/pages/About.js
```diff
@@ -149,6 +149,18 @@ const studentMembers = [
...
+        name: "Cindy Zheng",
+        photoSrc: Zheng_Vicky
+    }
+    { 
+        name: "Lisa Li-Liang",
+        photoSrc: Zheng_Vicky
+    }
+    {
+        name: "Kingston Lew",
+        photoSrc: Lew_Kingston
```
jeffreyshen19, line 163: This block is breaking the JS linter, make sure you include commas after each JSON object.

jeffreyshen19 (changes requested):
Looks good overall, and great work! Just a few small notes:
- I've added a comment, one of your files is failing the JS linter but it should be a 3 second fix.
- It looks like this PR is using the same branch as your first PR branch. In general, try creating new branches for every new feature (i.e. a new branch per PR). This will keep the Git history a bit cleaner.

ryaanahmed:
thanks folks -- there was a little merge screwiness that happened on this branch, but I've cleaned that up and merged.

PR #238, opened 2022-10-14
replace local dir to photos with aws_s3 dir

lisali-liang:
photo display and links are correct in admin photos category


jeffreyshen19 (dismissed):
Looks good, and nice work! Just one small quick fix, to keep our git history clean.

ryaanahmed:
Hey - went ahead and made the changes myself so we can get this in. @jeffreyshen19's comment is right on -- I'll add an extra settings file and speak with the admin page team about how to use it.

PR #239, opened 2022-10-14
Add logo to navigation bar

suzannej1:
Created a logo in adobe illustrator, exported as svg , and uploaded to assets folder. Added logo to navbar.html and sized logo to height 80 and width 80. 

image


backend/templates/includes/navbar.html
```diff
@@ -41,4 +43,11 @@
             </ul>
         </div>
     </div>
+    <!-- <img src="../includes/nav_logo.svg" height='50'/>  -->
```
jeffreyshen19, line 46: It's good practice to delete dead/commented out code if it's no longer useful, just to keep the files clean.

frontend/components/Logo.js
```diff
@@ -0,0 +1,49 @@
+// import React from "react";
```
jeffreyshen19, line 1: Same here, I think this file is no longer being used (double check!), so you can get rid of it so there's no confusion for future devs.

jeffreyshen19: And I think once you get rid of this file, you can also get rid of the logo.scss file below.

jeffreyshen19 (commented):
Great work, and I'm glad the logo finally worked (sorry about how complicated this all was)! My only comment is that there's some dead code/no longer used files that it's good practice to get rid of before submitting PRs. All good to merge, but would recommend taking care of them for the next PR!

PR #240, opened 2022-10-18
Map Square Photographers

maggiehuiliyao:
Implemented the photographers section of buttons of the map square. As a note, in titleFunc, `Map Square: ${photographer["map_square"]}`, it doesn't return the map square due to the fact that the data is missing. Only square 120 has data on which photographers were in the map square due to creating fake test data.

### DIFF
--- a/frontend/components/PhotoViewer.js
+++ b/frontend/components/PhotoViewer.js
@@ -1,7 +1,7 @@
 import React from "react";
 import * as PropTypes from "prop-types";
 
-function getValue(dictionary, key, default_val) {
+export function getValue(dictionary, key, default_val) {
     let result = dictionary[key];
     if (typeof result === "undefined") {
         result = default_val;
--- a/frontend/scss/styles.scss
+++ b/frontend/scss/styles.scss
@@ -158,7 +158,6 @@ footer {
     @extend .btn-secondary;
     border: .08em solid $color-blue-light;
     &:hover, &:focus {
-        color: white !important;
         background-color: $color-blue-light !important;
     }
 }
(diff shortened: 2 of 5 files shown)


frontend/scss/styles.scss
```diff
@@ -153,3 +153,12 @@ footer {
         background-color: $color-red !important;
     }
 }
+
+.btn-secondary-blue {
+    @extend .btn-secondary;
+    border: .08em solid $color-blue-light;
+    &:hover, &:focus {
+        color: white !important;
```
jeffreyshen19, line 161: When using @extend, it's good practice to only add styles that are different. Since `color: white !important` is already defined in the parent `.btn-secondary`, you can remove this line.

frontend/components/PhotoViewer.js
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import * as PropTypes from "prop-types";
 
-function getValue(dictionary, key, default_val) {
+export function getValue(dictionary, key, default_val) {
```
jeffreyshen19, line 4: It looks like this function is getting a key from a dictionary and, if undefined, returning a default value. This can actually be done using the or operator. For instance, instead of:

`x = getValue(dictionary, key, default_val)`

You can write:

`x = dictionary[key] || default_val`

See: https://stackoverflow.com/questions/9003999/how-to-get-value-from-object-with-default-value. No need to change all mentions of `getValue`, just a note for future reference.

jeffreyshen19 (commented):
Looks good, and great work! One small comment: in the future, when making frontend PRs/style changes, it's helpful to include a screenshot to see what it looks like.

jeffreyshen19 (approved):
Thanks for making the changes, merging in!

PR #247, opened 2022-10-28
Vivian jeffery barima nisha

vchinoda:
* Fixed random images on home page to zoom in less and fit in the background size as best as possible.
* Increased row height to allow top random images to display and the map's color to pop out.

* Shortcoming: Car image and top random images are overlapping

### DIFF
--- a/backend/config/settings/dev.py
+++ b/backend/config/settings/dev.py
@@ -22,14 +22,24 @@ pw_path = Path(PROJECT_ROOT, 'db_password.txt')
 with open(pw_path, 'r', encoding='utf-8') as pw_file:
     db_pw = pw_file.readline().strip()
 
+# DATABASES = {
+#     'default': {
+#        'ENGINE': 'django.db.backends.postgresql',
+#        'HOST': 'paris1970-urop-fa22.crdpmszp71qh.us-east-1.rds.amazonaws.com',
+#        'USER': 'urop',
+#        'NAME': 'paris1970-urop-fa22',
+#        'PORT': '5432',
+#        'PASSWORD': db_pw,
+#    }
+#}
+
+# NOTE(ra): AWS RDS is down on 2022-09-09 during our first hack session,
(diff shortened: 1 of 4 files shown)


frontend/pages/Home.js
```diff
@@ -11,11 +12,22 @@ export class HomePage extends React.Component {
 
     render() {
         return (
+            // <StrictMode>
+            //     <Collage/>
```
jeffreyshen19, line 16: It's generally good practice to remove dead code before submitting a PR. You can also remove the import  for `StrictMode` on line 1, as it's no longer needed.

frontend/components/collagephotos.js
```diff
@@ -0,0 +1,99 @@
...
+    var width = (window.screen.width*window.devicePixelRatio)/4;
+	var height = (window.screen.height*window.devicePixelRatio)/9;
+}
+
+
+
+// const width = (window.screen.width*window.devicePixelRatio)/4;
+// const height = (window.screen.height*window.devicePixelRatio)/6;
+
+export const images = [
```
jeffreyshen19, line 34: It's good practice to reduce redundancy in your code. Instead of copying the `width` and `height` for each image, you can build this list of JSON objects programmatically.

jeffreyshen19 (commented):
Looks good! I've made some comments on redundancy in your code, and there are also some warnings raised by the JavaScript linters that would be great to fix (either for this PR or the next one). A few small notes:

- Try including screenshots in the PR when making design changes 
- It looks like there are several different features going on in this branch. In general it's good to make branches and PRs for each individual feature rather than one big PR for your team. 

Great work though!

ryaanahmed:
nice work folks -- getting there. @Nisha-Nathan in particular note that I've made some changes factoring collagephotos.js and the main component for the collage together (since that didn't seem to need to spread across two files) and replacing a bit of redundant data with code per @jeffreyshen19's comment above. merging now.

PR #248, opened 2022-10-28
Adding pagination to photographers search

Menamonmon:
* Added scrollable pagination for the photographers list. (As you scroll to the end of the page, it loads the next batch of photographers. Currently, displaying 50 per batch)
* Added multiple fields for search (name, beginning of name, location, map square) and sorting with all of these fields

Future goals:
* Add a reset button that clears all of the search options to improve user experience if they're search queries don't have any matching results
* Changing the implementation of the `nameStartsWith` field to include option for multiple letters

Picture of how the page currently looks like
Screen Shot 2022-11-04 at 4 56 53 PM


Menamonmon:
@ryaanahmed I'm running into an error with the backend test for this PR. Locally, it's working completely fine. I might be completely wrong, but I think the database on the github backend tests is outdates or needs to have some migrations run on it. Please let me know when you approve this PR, so that @fangkelly can merge these changes into her ongoing PR for adding the future goals that are outlined in the PR description.

ryaanahmed:
> @ryaanahmed I'm running into an error with the backend test for this PR. Locally, it's working completely fine. I might be completely wrong, but I think the database on the github backend tests is outdates or needs to have some migrations run on it. Please let me know when you approve this PR, so that @fangkelly can merge these changes into her ongoing PR for adding the future goals that are outlined in the PR description.

Weird -- I'll take a look. As part of the testing process, we rerun the database migrations to setup the testing database... which is working just fine on the `main` branch, and your PR makes no changes that should affect that! I'll dig into it and see what's wrong.

Menamonmon:
hey @ryaanahmed, I think the error might be related to the pagination code that I added in this PR. Are there any restrictions on the use of the Pagination class in our backend code?

ryaanahmed:
> hey @ryaanahmed, I think the error might be related to the pagination code that I added in this PR. Are there any restrictions on the use of the Pagination class in our backend code?

Sorry, just now looking at this. Nope, that's not the problem. The problem is the code that's in `backend.app.views` but not encapsulated within a function.

Any code that's in a Python module but not wrapped inside a function gets executed when that file is imported, so it's generally an unsafe practice to leave any code in there unless you know it's safe to run under all circumstances in which the file will be imported.

What's happening here is that we run `python manage.py migrate` to setup the testing database. When `manage.py` is run, it imports the whole project, and at that import time it tries to run all that code, including this line:

```
set(Photographer.objects.all().values_list('approx_loc', flat=True))
```

... but there's no database yet! So that's why it's crashing on the CI server. I'll go ahead and fix the issue, and feel free to come ask me in the hack session tomorrow if you don't understand the explanation above. But in general, moving forward, make sure to stick your code in a function. Even for a script that you only execute yourself locally, it's best to stick stuff in a `main()` function and then use the `if __name__ == '__main__'` thing to call it. (See https://realpython.com/if-name-main-python/)

cc @jeffreyshen19

backend/app/views.py
```diff
@@ -112,24 +114,129 @@ def search_photographers(request):
...
+    res = Response({
+        "page_number": page_number,
+        "results": serialized_photographers.data,
+        "is_last_page": not current_page.has_next()
+    })
     return res
 
 
+@api_view(['GET'])
+def get_search_photographers_dropdown_options(request):
```
ryaanahmed, line 193: For now, this is good. Ultimately, since the list of locations is static and not going to change, this is something that we might want to remove from code, turn into data, and save in the frontend to avoid an additional backend query to make the photographers page.

backend/app/views.py
```diff
@@ -112,24 +114,129 @@ def search_photographers(request):
     TODO: Add pagination for both cases (when given a search query and when nothing is given) 
     so that the user is sent the first 50 results and they can view more results as they scroll down the page.
     """
+
+    def parse_order_by(order_by):
```
ryaanahmed, line 118: There's a lot going on here, but the structure is super clear! Nice job.

backend/app/views.py
```diff
@@ -112,24 +114,129 @@ def search_photographers(request):
...
+                set(Photographer.objects.all().values_list('approx_loc', flat=True))
+            )
+        )
+    )
+
+    squares = sorted(
+        filter(
+            lambda x: x is not None, 
+            list(
+                set(Photographer.objects.all().values_list('map_square_id', flat=True))
```
ryaanahmed, line 214: We should be sending map square numbers, not ids here. We shouldn't rely on the ID and map square number being the same.

frontend/pages/views/PhotographerListView.js
```diff
@@ -192,78 +260,123 @@ export class PhotographerListView extends React.Component {
...
-                        <ul className="list-inline">{this.getPhotoList()}</ul>
-                    </div>
+                <div className="photographerGallery">
+                    <ul className="list-inline" onScroll={this.handleScroll}>
+                        {this.getPhotoList()}
+                        <div className="photographers-results-footer">
+							{this.state.isLastPage
+								? (this.state.photographers.length === 0
+									? "No photographers found that match your search query!"
+									: "End of Results!!!")
```
ryaanahmed, line 375: Generally, don't leave temporary copy like this in the frontend. You want it to look professional... especially if you forget to change it later!

frontend/pages/views/PhotographerListView.js
```diff
@@ -83,30 +106,72 @@ export class PhotographerListView extends React.Component {
...
+            if (this.state.isLastPage) {
+                return;
+            }
+            const {name, nameStartsWith, location, square} = searchQueries;
+            const newPageNumber = this.state.pageNumber + 1;
             try {
-                const res = await fetch(`/api/search_photographers?name=${sq}`);
+                const res = await fetch(
+                    `/api/search_photographers?
+						name=${!name ? "" : name}&
```
ryaanahmed, line 147: The default values here are already empty strings (in the constructor), so no need to set empty strings here -- just use the values directly.

backend/app/views.py
```diff
@@ -112,24 +114,129 @@ def search_photographers(request):
...
+    # Pagination params
+    page_number = request.GET.get("page", None)
+    count_per_page = 50 
+
+    search_params = {}
+    
+    # Validating and adding all of the params
+    if name is not None and name.strip() != "":
+        search_params["name__icontains"] = name
+    if location is not None and location.strip() != "" and location in locations:
```
ryaanahmed, line 150: A bunch of this validation is a little over the top -- e.g., we're sending the locations list to the frontend and populating the request only from those locations, so we don't really need guardrails checking the value is still in locations. (Same with nameStartsWith, etc.)

ryaanahmed (approved):
Nice work overall! I've provided a bit of feedback within, but I know this whole page is in progress, and there are no major architectural red flags here, so keep going! Let's chat about sensible default values in hack session.

PR #250, opened 2022-10-28
Fix slow map page

D4S44K:
Ryaan walked me through how to make the map page load faster.

The main issue was that a new SQL query was being made every time a new photo was loaded for every map square. In order to resolve this, we used prefetch_related so that we were able to:
1) return all map squares with one database query, and
2) return all of the photos associated with a map square in a single query using ForeignKey.

Another optimization that Ryaan demonstrated was using the .count() method to count the number of photos rather than using len(), which requires us to loop through the photos twice.


ryaanahmed (approved):
Thanks! And thanks for removing that logging code before making the PR. Merging!

PR #252, opened 2022-11-01
Add photo map to tag view (in progress)

irinazoccolini:
So far, this PR adds a map to the tag view page. The map displays how many photos are found for the tag in each map square. 

I'm currently trying to figure out what should happen when a user clicks on a map square. Should it only state the number of photos for the tag in the map square, or should it also link to a page that shows all the photos with the tag for the map square? Right now, it does the same thing as the main map (links to all the photos in the map square). 

Screen Shot 2022-11-01 at 2 11 38 PM


frontend/pages/TagView.js
```diff
@@ -1,14 +1,181 @@
...
+                const numberOfPhotos = mapSquareData.num_photos;
+
+                const mapSquareBounds = [
+                    [coords.lat, coords.lng],
+                    [coords.lat - MAPSQUARE_HEIGHT, coords.lng - MAPSQUARE_WIDTH]
+                ];
+                const link = "/map_square/" + index;
+                let mapSquareBucket = "";
+                // set of conditionals to calculate photo density for heat map
+                if (numberOfPhotos > 0 && numberOfPhotos <= twentyPctMax) {
```
jeffreyshen19, line 34: These buckets can definitely be implemented automatically: instead of using multiple `if/else` blocks, you can use division to map numberOfPhotos to bucket #. This has a few benefits: (1) it's cleaner, and (2) is easier to change in the future i.e. if you want to change buckets from 5 to 10.

frontend/pages/TagView.js
```diff
@@ -1,14 +1,181 @@
...
+                    mapSquareBucket = "map-square box-five";
+                }
+
+                return (
+                    <Rectangle
+                        className={numberOfPhotos === 0 ? "map-grid" : mapSquareBucket}
+                        key={index}
+                        bounds={mapSquareBounds}>
+                        {// Shows map square numbers on each map square
+                            // <Marker position={L.polygon(mapSquareBounds).getBounds().getCenter()}
```
jeffreyshen19, line 52: It's generally good practice to remove dead code before submitting PRs

frontend/pages/TagView.js
```diff
@@ -1,14 +1,181 @@
...
         super(props);
+
+        this.state = {
+            mapData: null,
+            geojsonData: null,
+            filledMapSquares: null,
+            filledMapSquaresData: null,
+            isLgViewportUp: null,
+            mapSquare: null,
+            mapLat: 48.858859,
```
jeffreyshen19, line 92: it looks like these coordinates are used across the project, maybe it's worth adding a constant in the `ParisMap` component to use instead of copying/pasting every time.

frontend/pages/TagView.js
```diff
@@ -1,14 +1,181 @@
...
+                // it pulls those coordinates and makes those the coordinates of the marker
+                // Coords must be in (lat, lng)
+
+                // If the map square does not have the coordinates it sets them to (0, 0)
+                // NOTE(ra): this no longer happens, so we can probably remove this safety check
+                if (roughCoords) {
+                    const roughCoordsList = roughCoords.split(", ");
+                    const lat = parseFloat(roughCoordsList[0]);
+                    const lng = parseFloat(roughCoordsList[1]);
+                    mapSquare.topLeftCoords = {
```
jeffreyshen19, line 119: Nitpick but no need to repeat this block in both the `if` and `else` conditions: 

```
mapSquare.topLeftCoords = {
     lat,
     lng
};
```

jeffreyshen19 (changes requested):
Looks really good! I've made a few small style comments, mostly to reduce redundancy in the code. These should be pretty quick changes, but definitely feel free to move on or address them in your next PR—just lmk!

ryaanahmed:
> Right now, it does the same thing as the main map (links to all the photos in the map square).

@irinazoccolini -- I think this is the most sensible thing to do for now. Building a separate page that has the intersection of a map square and a tag probably isn't worth it. Maybe if we had a filtering mechanism for an individual map square...

ryaanahmed:
@jeffreyshen19's comments here are right on, but since they can be addressed in follow on work, I'm going ahead and merging this in. thanks!

PR #256, opened 2022-11-03
update Photo model to incorporate folders

### LINKED ISSUE #246: Move photo url generation to the backend as part of the Photo model
Right now, the frontend tends to construct URLs for individual photos. We do this by passing the `photoDir` in `app.views.render_view` and in `map_page` etc.

We should shove all of this back into Photo, so that the frontend can just do `photo.front_url` and `photo.slide_url` (or some other similar structure: we may want to have a dict that holds many urls for different resolutions, e.g.,) instead, and we don't have to expose the photos dir to the frontend at all.

ryaanahmed:
This PR updates the Photo model to accommodate the format of the final dataset. Specifically:
- Adds a `folder` field to `Photo`, which joins `number` and `map_square` in `unique_together`.
- Wherever we previously looked for photos based on `map_square` and `number`, we now add `folder`, since there can be multiple photos with the same number in a map square but in different subfolders
- Updated `integrate_new_images` to do so from the final dataset. It no longer renames files in place, as we're planning on retaining the original filenames.
- Pushes all photo/slide image url generation to the backend (Fixes #246)

Probably shouldn't be in this PR but it was helpful for this work:
- Adds `dev_local_db` and `dev_admin_db` settings files to make it easier for folks to either use a local sqlite3 db or login to the remote db with admin privileges. (@czheng10 -- take a look at this and ask me or @jeffreyshen19 if you need help using it for your team).


backend/app/models.py
```diff
@@ -21,14 +21,31 @@ class Photo(models.Model):
...
+    # The next pair (01_003 and 01_004) become photo 2, etc.
+    number = models.IntegerField()
+
+    # Furthermore, many map squares contained multiple folders.
+    # These folders seemed meaningful -- they often represent boundary points between photographers.
+    folder = models.IntegerField()
+
+    # The original full filename of the SLIDE as provided to us by 
+    # e.g., "BHVP_PH_CetaitParis_DP_0031_01_001"
+    shelfmark = models.CharField(max_length=252)
```
ryaanahmed, line 43: `integrate_new_images` doesn't currently save the shelfmark -- I totally forgot to do this. Let's not let this block the PR, I'll add this info in a follow-up PR.

backend/app/views.py
```diff
@@ -412,27 +417,11 @@ def get_photo_distances(request, photographer_num):
     return Response(sorted_photo_data)
 
 
-@api_view(['GET'])
```
ryaanahmed, line 415: This view was fully redundant with the get_map_square view.

.gitignore
```diff
@@ -153,6 +153,8 @@ yarn-debug.log*
 yarn-error.log*
 *.swp
 db_password.txt
+admin_db_password.txt
```
jeffreyshen19: This file name might be confusing, since there's already a file for the admin db password, you should rename so it's clear that it's for super-user privileges

jeffreyshen19 (approved):
Looks good, merging in!

PR #258, opened 2022-11-04
Add management command to import photo metadata

### LINKED ISSUE #257: Pull data from our new metadata spreadsheet into the remote db
We're loading the full real dataset now, so we need to add the metadata that's been transcribed by everyone in the map square entry spreadsheet (https://docs.google.com/spreadsheets/d/1gYgxReW67nddVOXhj3bDAd7yJX_YXEFTxkIyO9GpcCk/edit).

ryaanahmed:
Adds a management command to sync metadata exported from our map square entry Google Sheet into our database.

Fixes #257 

@jeffreyshen19 -- I already ran this command on the remote dev database, so photos for a given photographer are now available (if transcribed). But please do review before merging.


backend/app/management/commands/syncdb.py
```diff
@@ -1,518 +1,537 @@
 """
```
ryaanahmed, line 1: Not sure why this diff is so crazy: I commented out the entire file, and wrote a few paragraphs at the top explaining what it was and why we're keeping it around for a bit longer.

jeffreyshen19 (approved):

PR #259, opened 2022-11-15
Start cleaning up the analysis pipeline

ryaanahmed:
This PR starts cleaning up the analysis pipeline. It removes ideas that were around at the beginning of the project (e.g., that analyses might be run on models other than Photo) that we never developed, and have just sat around complicating the code.

It also updates the local image data retrieval that the analysis pipeline uses to use the current structure of the photos.

Tagging @jeffreyshen19 @JusticeV452 @irinazoccolini so you know that this is in, but I'm going to go ahead and self-merge once it passes tests.

### DIFF
--- a/backend/app/management/commands/runanalysis.py
+++ b/backend/app/management/commands/runanalysis.py
@@ -5,13 +5,11 @@ Syncs local db with data from project Google Sheet
 """
 
 import sys
-import os
 import json
 
 from importlib import import_module
 from typing import Callable
 
-from django.conf import settings
 from django.core.management.base import BaseCommand
 
 from app.common import print_header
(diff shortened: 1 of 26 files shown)


backend/app/management/commands/runanalysis.py
```diff
@@ -43,85 +49,51 @@ def handle(self, *args, **options):
             print_header('There is no analysis with that name.')
             sys.exit(1)
 
-        if analysis_module:
```
ryaanahmed, line 46: We used to save pickles of the analyses, so that we could restart analysis runs that were interrupted by a crash, but we haven't been using that feature, so I've removed it. We now do not overwrite existing analyses by default, but have a `--overwrite` flag to do so.

backend/app/models.py
```diff
@@ -51,58 +51,34 @@ class Photo(models.Model):
...
-    def has_valid_source(self, photo_dir=settings.LOCAL_PHOTOS_LOCATION):
-        return os.path.exists(os.path.join(photo_dir, f"{self.map_square.number}/"
-                                                      f"{self.number}_photo.jpg"))
-
-    def get_image_data(self, as_gray=False, use_pillow=False, src_dir=settings.LOCAL_PHOTOS_LOCATION):
+    def get_image_data(self, as_gray=False, use_pillow=False, photos_dir=settings.LOCAL_PHOTOS_DIR):
         """
         Get the image data via skimage's imread, for use in analyses
 
-        We try for a local filepath first, as that's faster,
```
ryaanahmed, line 62: Removed all of the Google drive fallback code, as it's now obsolete.

PR #261, opened 2022-11-15
Fixing the broken links on the photographer page

bukunmishodipo:
The previous implementation of the photographer's page was loading photos that were stored directly in the codebase, however we have a large number of photos. I updated 'urls.py' to add a file path that returns one photo from a given photographer and wrote the function 'get_photographer_photos' in 'views.py' that executes the function when it's called. 'PhotographerListView.js' was updated to use this file path for each photographer being loaded on the page.
Photos are currently not being generated due to the photos currently not being linked to photographers on the backend, but once that is sorted, things should be working.


backend/app/views.py
```diff
@@ -139,6 +139,17 @@ def get_photographer(request, photographer_number=None):
...
     return Response(serializer.data)
 
+@api_view(['GET'])
+def get_photographer_photos(request, photographer_number):
+    """
+    API endpoint to get a photographer based on the photographer_id
+    If given photographer_number, GETs associated photographs
+    """
+    photo_obj = Photo.objects.filter(photographer__number = photographer_number)
+    print(photo_obj)
```
ryaanahmed, line 149: Please remove debug print statements when making a PR.

backend/app/views.py
```diff
@@ -139,6 +139,17 @@ def get_photographer(request, photographer_number=None):
...
+    serializer = PhotoSerializer(photo_obj, many=photographer_number is None)
```
ryaanahmed, line 150: This validation is a little bit difficult to read, and that's a code smell that something is wrong. If photographer_number is None, then it doesn't make sense to return anything from this function, and it should probably send an error code that the frontend can interpret and display in some sensible way. You'd do this by checking whether `photographer_number is None` right at the top of this function, rather than inside of this serializer.

frontend/pages/views/PhotographerListView.js
```diff
@@ -99,7 +99,8 @@ export class PhotographerListView extends React.Component {
     }
 
     srcFunc(number) {
-        return `${this.props.photoListDir}/${number}_photo.jpg`;
+        // return `${this.props.photoListDir}/${number}_photo.jpg`;
```
ryaanahmed, line 102: Please delete old code like this rather than leaving in, as we don't need it anymore.

ryaanahmed (approved):
Looks great! I made a few comments, but I'll go ahead and fix those myself before merging.

ryaanahmed (changes requested):
Ah -- I wasn't quite right here. Although you've added an API endpoint that gets all of the photos for a given photographer, it isn't _quite_ wired up correctly in the frontend. For the sake of expediency, I'll make the changes, but please do take a look at the final code in this PR so you can see what I've changed.

PR #267, opened 2022-11-28
Modifying see below button

jeffreykwon:
- Added svg file for the two arrow icon for see below button
- Modifed js file so that see below button is now the two arrow icon
- tried to add see below text but am having trouble centering it
- made the button transparent


jeffreyshen19 (commented):
Looks good! I pushed a small change fixing an import error, and centered your see below text. However, the text placement still looks a bit off: 

Screen Shot 2022-11-29 at 4 44 54 PM

If you have some time, it would be great to fix the positioning—all good if not though, Ryaan or I could take care of it.

